PR #419, opened 2015-01-09
Moving the $json variable

poxtron:
If one adds an error though the ninja_forms_process action, then the $json variable will not be set and get the error : 
Undefined variable: json in ..../wp-content/plugins/ninja-forms/includes/display/processing/post-process.php 

According to the documentation one should be able to add errors on that action.


kstover:
Hey @poxtron, great catch! Can you re-submit this PR on the develop branch?

PR #422, opened 2015-01-13
Fix wp-cli plugin activate issue

saetia:
this will solve the missing nf_change_email_fav fatal error


saetia:
moved from master to develop with pr #423

PR #439, opened 2015-01-28
[Hotfix] Number field attributes space bugfix, which causes invalid name attr.

athlan:
This hotfix solves the problem that attributes in number field type are concatenated without space. This propagates the problem, that `name` parameter is invalid and browser do not send field's value and it is empty (`false`) while processing form submission - user cannot ubmit data from this field.

Please merge to master ASAP!


kstover:
Sorry that it took me so long to get to this; we've been super busy here!

PR #463, opened 2015-02-18
pass button text to button/submit types using 'value' key. closes #462

### LINKED ISSUE #462: Metabox output submit and button types need separate label and value parameters
I'm trying to move the Aweber extension settings into a tab with the rest of the NF settings. 

I went to create a submit type setting and this is what I am getting:

image

Essentially the label is serving as both the label and the button text. This seems redundant and it would be better to have a separate value parameter.

helgatheviking:
usage:

```
array(
	'name' => 'button',
	'type' => 'button',
	'value' => 'A button',
	'label' => 'And a label goes here',
),
```


helgatheviking:
I don't really know why your commits are in there but hopefully you can sort it out. Let me know if I pulled against the right branch this time.

PR #471, opened 2015-02-23
Added Polish .mo and .pl translation files

BTMPL:


kstover:
Hey. Thanks for the translation! I'm going to throw this into our Transifex.com account. That's where we're handling translations from now on. Also, nice Richard gravatar. :)

PR #484, opened 2015-02-27
Spanish Translation (almost complete)

eduardoarandah:
the most important parts translated to spanish


kstover:
Thanks for the pull request! We actually just had someone finish a complete Spanish translation, but thank you very much for the work!

PR #510, opened 2015-03-03
Bring add-on updater up to date

pippinsplugins:
The add-on updater class is quite a bit out of date. Version 1.6 is significantly better, more reliable, and makes add-on updates work on multisite.


pippinsplugins:
:+1:

PR #569, opened 2015-03-24
Date Time Field - issue/504

stuart-sequeira:
A new hidden field called Date Time that generates a time stamp upon form display.  Four options to format it.

Would love to add custom formatting but had a brain block and couldn't get it to work like the standard hidden field default values.


kstover:
Hey Stuart,

I really like this idea, but I'm going to try to hold off merging any new fields until June when the new field registration class is released.

PR #596, opened 2015-04-07
Conversion Reset Button - issue/564

### LINKED ISSUE #564: Conversion Reset
- [ ] Prioritize current forms over converted forms (if there is a conflict)
- [x] Add a conversion button to Forms -> Settings -> Advanced (delete conversion flags)

kjohnson:
Closes #564


kstover:
I think that we need to add a JS confirm box that will inform the user that they are about to re-attempt form conversion. This may end up giving them multiples of the same form if their previous conversion was successful. @kjohnson get with @wpnzach to hammer out what the wording of the prompt should be.

kjohnson:
Per conversation with @kstover, also need to add a description to the option in the settings.

kstover:
Doing some more testing, and I don't seem to be getting any fields or notifications in my forms when I run the conversion reset.

PR #610, opened 2015-04-13
Query String Default Values - issue/607

kstover:


kjohnson:
What about [Reserved Terms](http://codex.wordpress.org/Function_Reference/register_taxonomy#Reserved_Terms)? 

For example, "name" is a reserved term.

We might need to at least register some help text.

PR #620, opened 2015-04-15
Upgrade Handler - issue/609

### LINKED ISSUE #609: Upgrade Process Handler

kjohnson:
Closes #609


kjohnson:
Ready for testing.

kstover:
Ok. There are a couple of things that I noticed that need to be worked on before we merge:

1) When I updated my Ninja Forms version, I was automatically taken to the process screen for updates; we should only begin update processes as a result of user-interaction. It felt like everything was being hi-jacked.
2) I think that we should show the update list first with some more explanatory text. I know that Zach was supposed to write it for you, but I can tackle it since he is out of the office.
3) We should stay on the initial update list screen for a few seconds to give the reader time to take in the information.
4) When we return to the update list, wait a couple of seconds so that the user can see the items check off and has an idea of what's coming next.
5) I got a notice to run the update routines when I did a fresh installation. We need to make sure that the flags are being added on a new install so that the updater doesn't flag.
6) When the update routines are complete, we should show a message saying so and then provide a link to the all forms list or the about page.

kjohnson:
On a clean install, the upgrade submissions flag was not set.

kjohnson:
@kstover just needs to text updated

kjohnson:
Closing pull request to migrate newer upgrade handler.

PR #632, opened 2015-04-19
Fix #631

### LINKED ISSUE #631: Undefined offset preventing form from properly saving
When `WP_DEBUG` is set to true it shows an undefined offset which in turn prevents the forms from properly saving in the form editor. The loading icon continues to spin. I suspect that WP_DEBUG is showing the error which is breaking the ajax

    PHP Notice:  Undefined offset: 57 in /wp-content/plugins/ninja-forms/includes/admin/ajax.php on line 115

sumobi:


kstover:
Hey. Could you re-pull this against the master branch?

PR #643, opened 2015-04-21
add local currency seperator support

kjohnson:
Use the WordPress local localized script for stripping the `thousands separator` and converting the `decimal point`.


kstover:
Once we hear back from the user who was having this issue, we'll merge this. Good job.

PR #652, opened 2015-04-23
Update ninja-forms-de_DE.po

MarvinLudwig:
Changed msgid "Please answer the anti-spam question correctly." to reflect formal version "Sie" and added info about case sensitivity.


kstover:

PR #658, opened 2015-04-23
Set $display_style

### LINKED ISSUE #657: PHP Notice:  Undefined variable: display_style
When you are viewing a submission that has a list (dropdown) there's a PHP notice:

    PHP Notice:  Undefined variable: display_style in /wordpress/wp-content/plugins/ninja-forms/includes/fields/list.php on line 765

It only seems to affect the dropdown:

screen shot 2015-04-23 at 2 42 59 pm

sumobi:
Fixes #657

Although this fixes the PHP notice, I'm not seeing any `$display_style` in the `$option` array. It looks like this might have just been a left over variable after a copy and paste, correct me if I'm wrong. Removing it would be another option, let me know!

### DIFF
--- a/includes/fields/list.php
+++ b/includes/fields/list.php
@@ -725,6 +725,12 @@ function nf_field_list_edit_sub_value( $field_id, $user_value, $field ) {
 						$label = '';
 					}
 
+					if(isset($option['display_style'])){
+						$display_style = $option['display_style'];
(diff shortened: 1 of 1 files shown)


kstover:
We can leave your fix in. It was added so that extensions or other plugins can hide specific list options.

PR #660, opened 2015-04-23
Check that $data['label'] is set

sumobi:
Fixes #659 

Although this fixes the notice, I'm not seeing a `$data` array. Should we go ahead and remove it?

PR #672, opened 2015-04-30
Remove unused $field_label variables

sumobi:
Related: https://github.com/wpninjas/ninja-forms/pull/660#issuecomment-97470518


kstover:

PR #678, opened 2015-05-06
Added country South Sudan and German, Danish translations of it.

eriadam:
Added the country to the country.php, regenerated and extended German and Danish translation files.


kstover:
Could you re-submit without the additional language files? I can add you to our account at transifex.com, and we can get the translations in there. That's where we're handling translations from now on.

PR #683, opened 2015-05-11
Updated Dutch translations

martijngastkemper:


kstover:
Hey. Can I get you added to our transifex account so that you can update the lines that need to be updated? That's where we handle our translation stuff.

PR #692, opened 2015-05-19
NF Admin Modal JS will now output all required structural HTML

### LINKED ISSUE #691: NF Admin Modal Structural HTML
When using the NF Admin Modal, the structural HTML necessary should be added by the JS rather than expecting it to exist on the page already. This will make implementing it much easier.

kstover:
Closes #691


kjohnson:
JS needs to be minified.

PR #736, opened 2015-06-01
Remove unused function arguments for refresh_plugin_settings()

wpnzach:
removed arguments from refresh_plugin_settings


wpnzach:
closes 735

kjohnson:
Do these need to be kept for backwards compatibility? If this is used anywhere with arguments, then removing them from the function declaration will throw an error.

kstover:
It is referenced one other place in core, and I can change that.

kstover:
None of our extensions reference it.

PR #741, opened 2015-06-04
Response message display fix

bostondv:
I found that the response message logic to hide it was incorrect and causing the markup to be displayed even when there was no success or error message to display.

This PR also adds gulpfile command to compile the ninja-forms-display.js (which was missing) and a show() function call when updating error message and a hide() on before submit.


kstover:

PR #742, opened 2015-06-04
Removes wrapping div from response messages

bostondv:
In my tests there were no issue removing these divs. I've removed them to clean up the markup and because while trying to adapt ninja forms to work with Bootstrap these inner divs caused some spacing problems.


kstover:
@bostondv This was reverted because the div tag actually separates multiple errors. I'd be up for discussing a better solution, however.

bostondv:
Ok, so here's a thought: If one error, don't wrap it in anything. If multiple errors, put them in an unordered list. 

Multiple errors would then be nicely separated and both cases would work well with Bootstrap. We'd just need to add a bit to the Ninja Forms css to make sure the list margins look good within the alerts.

What do you think? I'm happy to amend the PR if we agree on a solution.

PR #823, opened 2015-08-04
Improving ninja-forms-fr_FR

epigeyre:
I have improved the french translation and corrected some errors in the file:

- Better translation,
- Improved punctuation,
- Correction of missing letters in `<strong>` tags,
- Use of French exemple instead of the American National Security Number.


kstover:

PR #891, opened 2015-08-26
Changed submission data temporary storage to wp_session rather than t…

### LINKED ISSUE #890: Transient Caching Submission Data
When we setup our transient for storing submission data (used in redirect actions, PayPal extension, etc.), it can easily be cached by shared hosts. This means that redirects that use shortcodes on the landing page and PayPal will fail. The solution is to store this submission data in our wp_session rather than a transient.

kstover:
…ransient. Closes #890.

### DIFF
--- a/classes/session.php
+++ b/classes/session.php
@@ -112,6 +112,18 @@ class NF_Session {
 	}
 
 	/**
+	 * Delete a session variable
+	 *
+	 * @since 2.9.28
+	 * @param string $key
+	 * @return void
+	 */
+	public function delete() {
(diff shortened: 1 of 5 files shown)


kjohnson:
## Tests

- [ ] Redirect (with a shortcode on the landing page)
- [ ] PayPal Transaction

kjohnson:
Both tests were successful.

klhall1987:
## Tests

- [] Redirect (with a shortcode on the landing page)
- [ ] PayPal Transaction

klhall1987:
Both test were successful.

PR #950, opened 2015-12-06
Simple tidy up

thehereward:


kjohnson:
Hey @thehereward, thank you for the PR.

This is a library that we may not be using for Ninja Forms 3.0.

However, if you are interested in contributing, we'd love to work with you. Just let me know :)

Also, checkout http://developer.ninjaforms.com/ and http://developer.ninjaforms.com/slack for development information.

PR #975, opened 2015-12-11
Issue/undefined var

roykho:
Fixes issue with undefined variable on post edit page.


kjohnson:
@roykho the variable `$view_post_link_html` will be removed, since the message updated is now applied only to the submission post type which does not have a front facing view.

roykho:
@kjohnson thanks!

PR #1006, opened 2015-12-21
Filter output on page reload

kjohnson:
Use PHP native functions to filter form field output.

### DIFF
--- a/includes/database.php
+++ b/includes/database.php
@@ -238,28 +238,8 @@ function nf_wp_kses_post_deep( $value ){
         array_map( 'nf_wp_kses_post_deep', $value ) :
         wp_kses_post($value);
 
-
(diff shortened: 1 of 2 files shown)


kjohnson:
@klhall1987 and @kjohnson have tested this fix with the two most recent XSS payloads. Also, the Email Action and Success Message actions were tested for malformed tags caused by previous fixes.

oligoform:
**+1 for better security!**
but: it breaks my forms. not all fields are shown,
e.g. the submit-button lacks the given field-option
and the checkbox give a *Illegal string offset 'default_value'*
maybe you´ve an idea? 
so far i´ve uncommented line 118/119 (filter_var) temporarily.
(wp 4.4, ninja forms 2.9.32, PHP 5.6.13)

PR #1108, opened 2016-02-09
add `ninja_forms_loaded` hook for extensions.

helgatheviking:
avoid fatal errors when extensions are active but NF is not


kjohnson:
Already added the action, but I did copy over the comment.

PR #1124, opened 2016-02-25
add translations for de_CH region

rafhun:
As I am using this plugin on many  websites base from Switzerland I decided to update the very outdated de_DE (german) translations and adjust them where necessary to the Swiss German standards.

Also I am rather new to the whole open source community and PR's and such, so forgive any possible mistakes I made, and help me get it right the next time.

I've reviewed the contributing guidelines but found it to be very difficult to impossible to create translations from the master branch, since that one is practically empty. So I went for the 2.9.x branch. Is there a better / preferred way to do this?


kjohnson:
Hey @rafhum,

WorsPress recently made the switch to handling plugin translation on WordPress.org. They are hosting a GlotPress install to manage these.

https://translate.wordpress.org/projects/wp-plugins/ninja-forms

Thank you for taking the time to work on these translations.

Issue #1164, opened 2016-03-04
Allow NF_Database_Migrations class to be extended by addons


polevaultweb:
Currently the private visibility of the `$migrations` property means that the class can't be extended.

PR #1165, opened 2016-03-04
Set property to protected so addons can extend migration class for custom tables

polevaultweb:
https://github.com/wpninjas/ninja-forms/issues/1164


kstover:
@polevaultweb Can you re-submit this PR to master?

PR #1166, opened 2016-03-04
Allow addons to use helpers

polevaultweb:
If an addon uses templates and config, this means they can use these methods instead of rolling own


polevaultweb:
Bit overkill

PR #1168, opened 2016-03-04
Allow defaults to be passed to get setting and defer update of db option

polevaultweb:


kstover:
@polevaultweb Can you re-submit this PR to master?

PR #1201, opened 2016-03-07
Switch to Summernote for sub limit / logged out msg

wpnzach:
Switch to Summernote for sub limit / logged out msg


kstover:
Fixed in 9ec00ae31d1f0afaaac23520021ab751ded11cd0.

PR #1258, opened 2016-03-18
Set property to protected so addons can extend migration class for custom tables

### LINKED ISSUE #1164: Allow NF_Database_Migrations class to be extended by addons

polevaultweb:
Resolves #1164

Supersedes https://github.com/wpninjas/ninja-forms/pull/1165


polevaultweb:
@kstover raised :)

PR #1259, opened 2016-03-18
Allow defaults to be passed to get setting and defer update of db option

polevaultweb:
Supersedes https://github.com/wpninjas/ninja-forms/pull/1168


polevaultweb:
@kstover raised :)

PR #1280, opened 2016-04-05
Make [ninja_form] text in table filterable again

barrykooij:


barrykooij:
This used to be filterable here: https://github.com/wpninjas/ninja-forms/blob/3.0/deprecated/includes/admin/pages/ninja-forms/tabs/form-list/form-list.php#L176

PR #1294, opened 2016-04-11
Adjusted builder to adapt to a full screen instead of within the WP w…

kstover:
…rapper.


kjohnson:
The blue menu button is hidden by the top bar on small screens.

PR #1318, opened 2016-04-21
Add filter to <select> open tag for easier HTML manipulation

CFXd:
I found the need for this when using a jQuery select library, specifically, I needed to add a `data-placeholder` HTML attribute and found the only way was via a really nasty PHP regex.

### DIFF
--- a/deprecated/includes/fields/list.php
+++ b/deprecated/includes/fields/list.php
@@ -265,14 +265,14 @@ function ninja_forms_field_list_display( $field_id, $data, $form_id = '' ){
 
 	switch($list_type){
 		case 'dropdown':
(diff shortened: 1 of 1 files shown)


kstover:
Hey @CFXd, data-attributes are something that we are going to be adding in 3.0. For now, I'm going to have @klhall1987 test out this PR to make sure everything works as expected, and then we'll flag it as ready to merge.

kjohnson:
@klhall1987 have you had a chance to test this?

klhall1987:
@kjohnson I have not. I'll Add this to my list of items to in the next few days.

klhall1987:
Tested and seems to be working correctly.

PR #1323, opened 2016-04-26
fix: get ninja forms compatible to pjax

eteubert:
Adds compatibility to https://github.com/MoOx/pjax.

The main idea here is to wrap the code in `ninja-forms-display.js` in a globally accessible function so it can be called whenever needed. If you prefer I could remove the "pjax:success" listener, because it can easily be moved into the theme that is using pjax.


kstover:
@klhall1987 I need you to test this PR to make sure that it doesn't break anything with submissions/form display. Once you pass it, we can flag it for merger.

klhall1987:
Tested form builder, form display, and form submission in both 2.9.x code base and 3.0 code base both appear to be stable.

saetia:
this should support turbolinks as well

saetia:
yep. this breaks ninjaforms for any site running turbolinks instead of pjax.
kill me now

kjohnson:
@saetia was Ninja Forms working with Turbolinks before this change?

saetia:
yep, with jquery.turbolinks.js.

TypeError: init_all_the_ninja_things is not a function. (In 'init_all_the_ninja_things()', 'init_all_the_ninja_things' is undefined)

saetia:
move the ```jquery(document).ready``` block below the ```var init_all_the_things``` declaration to fix this.

kjohnson:
@saetia Can you submit a new pull request?

saetia:
@kjohnson yep, i'll submit a PR after fixing some of the other issues this caused like exponential event binding on every ajax page load. the 'timed submit button' countdown ends up having multiple interval timers acting on it.

PR #1329, opened 2016-04-28
Fix a fatal error in the NF_AddonChecker class

fjarrett:
```
Fatal error: Call to undefined function get_plugins() in wp-content/plugins/ninja-forms/lib/NF_AddonChecker.php on line 13
```

This PR requires `wp-admin/includes/plugin.php` before attempting to call `get_plugins()` in the `NF_AddonChecker` class constructor.

/cc @kjohnson @kstover @jonathanbardo

### DIFF
--- a/lib/NF_AddonChecker.php
+++ b/lib/NF_AddonChecker.php
@@ -10,6 +10,12 @@ final class NF_AddonChecker
 
     public function check_plugins()
     {
+        if ( ! function_exists( 'get_plugins' ) ) {
+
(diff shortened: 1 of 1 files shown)


kjohnson:
@klhall1987 

With and without a non-THREE-compatible addon installed:

- [ ] Upgrade Submenu ( Forms -> Upgrade )
- [ ] Konami Code (All Forms Table)

klhall1987:
I've the conditions displayed here and everything looks to be performing correctly.

PR #1348, opened 2016-05-12
Fix for ajax page replacement error and side effects

saetia:
- Changed `init_all_the_ninja_things()` from a function expression to a declaration to avoid the call to an undefined function.
- Removed the `ajax:success` event binding. That binds `init_all_the_ninja_things()` to every single page and partial replacement, which includes pages that most likely don't have contact forms. It will also cause issues for people replacing an unrelated section of the ninja form'd page and not expecting the form to reinitialize. The developer refreshing the form section with pjax will know when to call the `init_all_the_ninja_things()` function.
- Removed the `jQuery(document).ready` wrapper around the `init_all_the_ninja_things()` call. It shouldn't be necessary because the javascript is included in the footer after the dom is ready. This prevents the `$.ready` binding from tagging along with every subsequent page load after hitting the ninja form'd page.
- Changed the timed submit button countdown timer to a global variable to prevent ajax page replacements from attaching multiple timers

#1323


saetia:
do the minified versions get automatically generated or should i have taken care of that?

saetia:
@eteubert if this gets accepted, it will affect your site

kjohnson:
@saetia you would need to re-minify the JS.

kjohnson:
@saetia how does this work with pjax? To be honest, I don't know much about either pjax or turbolinks, and need to do some research.

saetia:
it doesn't do anything special for pjax or turbolinks; this PR makes it agnostic.
turbolinks out-of-the-box replaces the entire body of a page through an ajax call when you click any link. pjax can do the same thing, but because it requires manual configuration of the div getting replaced it may be used for things like tabs or accordions. pjax is basically jQuery.load() with window.history api enhancements.

if someone is just updating a twitter feed with pjax, the way ninjaforms behaves right now is that it will run the init_all_the_ninja_things() afterward and reset the countdown timer on the ninja form. this would drive someone crazy figuring out the cause of that bug.

saetia:
If page content is loaded via turbolinks and it includes javascript tags inside the body (like a ninja forms page), that javascript will be evaluated on render. However, if someone were to replace only the content within the `<form>` tag they would need to call manually call `init_all_the_ninja_things()`.

eteubert:
Thanks for the heads up @saetia :) /cc @christophwolff

saetia:
@kjohnson anymore thought given to this? i'm eager to get this resolved before the next release so my clients don't update and break their sites again.

kjohnson:
@saetia I'm am working to get this RP in the next release. Are you available anytime today for a chat?

saetia:
great! yes, i’ll be available all day

kjohnson:
@saetia http://developer.ninjaforms.com/slack/

PR #1378, opened 2016-05-31
Translations didn't work anymore in deprecated mode

haet:
I added the parent directory to the $lang_dir so the correct directory for translation files is `ninja-forms/deprecated/lang/` instead of just `deprecated/lang/`


kjohnson:
LGTM

PR #1384, opened 2016-06-01
Add database check for THREE forms table.

kjohnson:
To alleviate unintentional upgrades to THREE, we need to check to see if the upgrade process was completed.

This change checks the database for the `nf3_forms` table. If the table is not found, then load deprecated.


klhall1987:
Tested and ready for merge.

PR #1388, opened 2016-06-02
Added help text on radio groups

iamntz:
Hey guys. 

I know this is deprecated, but until V3 is merged, I guess this should be a thing: allow custom `help_text` for each radio field.


kjohnson:
@iamntz do you have an example of this being used? I'm not seeing a standard field in core that uses a radio group in the field settings.

iamntz:
Sure thing! It's not any standard core field that is using this, but a custom field.

Before this patch, you only could have a `help_text` by a radio group basis. Which means that you define one `help_text` key, at the field level, then that help will be displayed on all radio items. Which it makes little sense.

http://img.iamntz.com/jing/video_2016-06-13__14_18.mp4

However, you can now specify `help_text` keys on an item basis:

```
'options' => [
	[
		'name' => __('Auto Approve'),
		'value' => 'auto',
		'help_text' => 'Foo'
	],

	[
		'name' => __('Admin Approve'),
		'value' => 'admin',
		'help_text' => 'Bar'
	],
```

http://img.iamntz.com/jing/video_2016-06-13__11_57.mp4

kjohnson:
Ok, cool. That makes sense. Just didn't have a use case.

Issue #1425, opened 2016-06-14
Add a filter to force loading of 2.x


Shelob9:
I recently ran into an issue where the site got opted into 3.0 beta and I couldn't get it back to loading 2.x without hacking the plugin, which worked but was not desirable.

A filter that would force it to load the old code would have been really helpful. I will submit a PR with this filter.

kstover:
This shouldn't be an issue with version 3.0.x being the official version. If you still need this taken care of, you can re-open and we'll talk about it.

PR #1426, opened 2016-06-14
Add a filter to force loading of 2.x

Shelob9:
#1425 

This PR introduces `ninja_forms_force_load_deprecated` filter with a default value of false. It would allow for forcing loading of 2.X code with `add_filter( 'ninja_forms_force_load_deprecated', '__return_true' );`

### DIFF
--- a/ninja-forms.php
+++ b/ninja-forms.php
@@ -24,7 +24,7 @@ if( version_compare( get_option( 'ninja_forms_version', '0.0.0' ), '3', '<' ) ||
     update_option( 'ninja_forms_load_deprecated', TRUE );
 }
 
-if( get_option( 'ninja_forms_load_deprecated', FALSE ) && ! ( isset( $_POST[ 'nf2to3' ] ) && ( defined( 'DOING_AJAX' ) && DOING_AJAX ) ) ){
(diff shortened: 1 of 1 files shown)


kjohnson:
@Shelob9 You can rollback by going to Forms -> Settings -> Advanced -> "Rollback"

Also, if you define `NF_DEV` as `TRUE`, similar to `WP_DEBUG`, then you will see a switcher in the admin bar.

See: http://developer.ninjaforms.com/codex/version-switching/

kjohnson:
I had a chance to talk to @wpnzach about your use case, @Shelob9.

This should no longer be necessary, as we have a fallback in place for unintentional THREE upgrades.

Also, any version forcing that does not also consider the upgrade process would create a completely broken user experience.

I would rather not leave our plugin so vulnerable as to be shutdown by an arbitrary filter, but would rather protect against illegitimate upgrades (which we have already done).

Shelob9:
@kjohnson in our situation we did not have the ability to use the admin since the site was down and changing the flags in the database didn't help, also it was probably the fault of some stupid stuff a ThemeForest theme was doing.

What would be great about having the filter is I could commit the callback in a mu-plugin and then not worry about having to re-hack the main plugin file next time their is a Ninja Forms update.

jameslaws:
I think Kyles point is that shouldn't be able to happen again because we not only check that option but whether the new tables were created for 3.0. The tables are only created when a user explicitly chooses to go through the upgrade process and when that process was presented because they actually qualify for the upgrade. 

A filter means that anyone, for good or poor intent, could prevent a user from ever being able to upgrade to 3.0. That's not an ideal solution.

PR #1432, opened 2016-06-16
Render.php / NF_Display_Render class refactor

iamntz:
I'm doing some refactorings on this class because it have some huge methods and some code duplication.

For some reasons, the display method is called... localize. Any particular reasons on that? Or i'm free to rename the method (and leave the old name in place, for compatibility reasons)?

### DIFF
--- a/includes/Display/Render.php
+++ b/includes/Display/Render.php
@@ -48,7 +48,7 @@ final class NF_Display_Render
         }
     }
 
-    public static function form_has_max_submission( $form )
+    private static function form_has_max_submission( $form )
     {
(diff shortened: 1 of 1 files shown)


iamntz:
Alright, so the first stage is complete. I refactored the `localize` method by extracting most of the tuff into small, contained methods. There is some duplication inside the preview method, but that's a subject for another PR.

Also, this contains a tweak on submission counter when a form has a max allowed submissions.

kstover:
@kjohnson Need input on this, please.

kjohnson:
@kstover generally looks good to me, but I'll need to take a closer look.

klhall1987:

PR #1436, opened 2016-06-17
Wrap input-limit init in document ready callback.

kjohnson:


klhall1987:
Tested, and working correctly.

PR #1437, opened 2016-06-19
Removed untranslatable string

iamntz:
Get text doesn't parse PHP, so the removed part do exactly nothing in the translation process.


kjohnson:
Ah, yup. Straggler. Thanks!

klhall1987:
Tested and working properly. Passing to @kstover for merge.

PR #1452, opened 2016-06-21
Changed the System Status so that two items won't display as bytes.

### LINKED ISSUE #1321: System Status in 3.0 Showing B instead of MB
The 2 places that it is showing incorrectly are:
WP Memory Limit: 40 B
PHP Post Max Size: 32 B

jmcelhaney:
Changed the System Status so that WP Memory Limit and PHP Post Max Size won't display as bytes. Closes #1321

Recreation steps.
System Status under 3.0 should no longer show sizes in bytes for WP Memory Limit and PHP Post Max Size.


kstover:
@klhall1987 We need to test and make sure that this is good. No reason to delay such a minor change. :P

klhall1987:
@kstover Tested and ready for merge

PR #1453, opened 2016-06-21
Add debug response output.

kjohnson:
@klhall1987 You can use this to test the debug response for addons.


klhall1987:
Tested and working correctly.

PR #1494, opened 2016-07-12
Fix initial required check for list fields

### LINKED ISSUE #1482: Required multi-select and checkbox list fields are breaking on the front end in 3.0
In the 3.0 code base, if a form has a multi-select or checkbox list field that is set to be required and does not have a default value selected then, when the form is viewed on the front-end end the multi-select field and everything after isn't being displayed. There is a JS error appearing in the console. I've detailed the error in the screen shot below. 

Screenshot: 
screen shot 2016-07-06 at 2 00 04 pm

Replication Steps(in the 3.0 codebase):
1. Create a form.
2. Add a multi-select field and submit button.
3. Edit the multi-select and make it a required field. 
4. Preview the form.

kjohnson:
Closes #1482


kjohnson:
Ok, so, we really need to fix this gulp/re-gulp thing.

klhall1987:
The Multi-select/checkbox list required fields with no default values are no longer breaking display on the front end; however, if a multi-select or checkbox field is present the form will no long submit. I've take a screenshot of what I'm seeing in the network tab in the admin-ajax.php response.   

Screenshot
screen shot 2016-07-19 at 11 39 27 am

kjohnson:
@klhall1987 back to you for testing.

klhall1987:
Tested and working correctly passing to @kstover for merge.

PR #1505, opened 2016-07-15
Security fix 19671

klhall1987:
This is a security fix for the issue detailed in the ticket linked below. 

https://wpninjas.desk.com/web/agent/case/19671

### DIFF
--- a/deprecated/classes/download-all-subs.php
+++ b/deprecated/classes/download-all-subs.php
@@ -2,6 +2,8 @@
 
 class NF_Download_All_Subs extends NF_Step_Processing {
 
+	public $test = false;
+
 	function __construct() {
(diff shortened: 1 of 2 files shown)


deprecated/classes/download-all-subs.php
```diff
@@ -2,21 +2,23 @@
 
 class NF_Download_All_Subs extends NF_Step_Processing {
 
+	public $loaded = false;
```
kjohnson: What is the purpose of this property? I see that it is set in the `loading()` method, but I don't see it being checked or otherwise referenced.

kjohnson:
@kstover Credit `SumOfPwn.nl` in the changelog

PR #1517, opened 2016-07-18
Hide form fields when re-starting a submission.

kjohnson:
When re-starting a form submission hide form fields (but not the submit field).


kjohnson:
@klhall1987 ready to test with wpninjas/ninja-forms-paypal-express, as that is the only add-on that currently halts and re-starts form submission.

klhall1987:
Tested and working correctly, ready for merge.

PR #1519, opened 2016-07-18
filter checkbox checked/unchecked calc values for merge tags.

### LINKED ISSUE #1511: Single checkbox fields sending wrong amount to payment gateways
In 3.0, single checkbox fields are not sending their checked or not checked values to Payment Gateways(in this case PayPal). Instead it's sending either a 1 or a 0 as the value. I've detailed the value in the screencast below. 

Screencast: 
https://goo.gl/5gw9fM

Replication Steps(In 3.0):
1. Create a new form.
2. Add two single checkbox fields and give them both uncheck and check calc values. 
3. Goto Advanced tab and create a calculation that adds the two single checkbox fields together. 
4. Go to the email and actions tab and create a Collect Payment action. 
5. Set the Collect Payment action to use PayPal and set the Payment total to the calculation variable you just made. 
6. Preview the form check one of the check boxes and leave the other other unchecked. When you submit the form. You'll be taken to Paypal with a $1.00 total.

kjohnson:
Closes #1511 .

### DIFF
--- a/includes/Fields/Checkbox.php
+++ b/includes/Fields/Checkbox.php
@@ -32,6 +32,8 @@ class NF_Fields_Checkbox extends NF_Abstracts_Input
         $this->_settings[ 'label_pos' ][ 'value' ] = 'right';
 
         add_filter( 'ninja_forms_custom_columns', array( $this, 'custom_columns' ), 10, 2 );
+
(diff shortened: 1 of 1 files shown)


klhall1987:
Tested and working correctly, passing to @kstover for merge.

PR #1524, opened 2016-07-22
Removed disable input setting from the following fields: Product, shipping, and total Fields

### LINKED ISSUE #1512: In 3.0 remove disable input setting from shipping and total field
Currently in 3.0, both shipping and total fields a have disable input setting. These fields do not have a way to input data on the front end. I think I would be a good idea to remove these settings from the fields. 

Screenshots:
Shipping:
http://goo.gl/uftCn1

Total: 
http://goo.gl/M9NwMP

klhall1987:
I removed the disable input setting form the product field, shipping field, and the total field. Closes #1512


kjohnson:
@klhall1987 @kstover should we also remove the `required` setting from `shipping` and `total`, as these fields are not for user input?

klhall1987:
Tested and working correctly. Sending to @kstover for merge.

PR #1526, opened 2016-07-22
Removed disable input and input limit on the credit card fields. 

### LINKED ISSUE #1525: Remove restriction settings from credit card field in 3.0
Remove the restriction settings from the credit card field in 3.0

klhall1987:
Removed disable input and input limit on the credit card fields. Closes #1525

### DIFF
--- a/includes/Fields/CreditCard.php
+++ b/includes/Fields/CreditCard.php
@@ -17,6 +17,8 @@ class NF_Fields_CreditCard extends NF_Abstracts_Field
 
     protected $_test_value = '';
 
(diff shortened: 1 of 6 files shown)


kjohnson:
@klhall1987 The `CreditCardZip` field still has an `input_limit_set` setting.

klhall1987:
@kjohnson I thought users might need to set an input limit on the creditcardzip field. Or can they just use an input mask to do this?

kjohnson:
@klhall1987 My guess is zip codes would be an input mask, if anything. Also, different areas have different zip code formats.

Any thoughts, @kstover ?

kstover:
I'd say it has an input mask.

PR #1580, opened 2016-08-08
Additional Options for Default Value Selection

chadfranklin:
Added post_author_id, post_type, and custom_post_meta as options in the default value selection of the textbox and hidden fields. Added post_author_id to the number field. This contribution is at the request of a recent support ticket (case no. 20782).


kjohnson:
@chadfranklin this is something that we would probably roll into a 3.0 feature, as opposed in introducing it in the deprecated codebase.

PR #1586, opened 2016-08-10
Fix Textarea Encoding on Saved Submission Load

clifgriffin:
When loading a saved submission, single and double quotes are encoded into HTML entities. 

This modifies the filter_var statement responsible to exclude quotes from the sanitization.


klhall1987:
Tested and seems to be working passing to @kjohnson for merge.

PR #1591, opened 2016-08-10
Remove unnecessary transient setter.

kjohnson:
Possibly a copy and paste error during code migration?


klhall1987:
Tested, passing to @kstover for merge.

PR #1593, opened 2016-08-10
Calculation Conversion

kjohnson:


includes/Actions/Email.php
```diff
@@ -85,7 +85,7 @@ private function _get_attachments( $settings, $data )
     {
         $attachments = array();
 
-        if( $settings[ 'attach_csv' ] ){
+        if( 1 == $settings[ 'attach_csv' ] ){
```
kjohnson: This change was accidentally made in the branch, before being ported to another.

includes/Database/Models/Form.php
```diff
@@ -361,44 +361,6 @@ public function import_field_backwards_compatibility( $field )
...
-                if ('auto' == $field['calc_method'] && isset( $field['calc'] ) ) {
-                    // TODO: Support 'auto-totaling' calculations.
-                }
-            }
-
-            unset( $field[ 'calc' ] );
-            unset( $field[ 'calc_eq' ] );
-            unset( $field[ 'calc_method' ] );
-        }
-
```
kjohnson: Removed the old conversion to a note field, since we are now converting the calculations themselves.

lib/NF_Conversion.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+interface NF_Conversion
+{
+    public function run();
+}
```
kjohnson: Created a `conversion` interface for future conversion.

ninja-forms.php
```diff
@@ -14,6 +14,8 @@
 
 require_once dirname( __FILE__ ) . '/lib/NF_VersionSwitcher.php';
 require_once dirname( __FILE__ ) . '/lib/NF_Tracking.php';
+require_once dirname( __FILE__ ) . '/lib/NF_Conversion.php';
+require_once dirname( __FILE__ ) . '/lib/Conversion/Calculations.php';
```
kjohnson: Should eventually extract all conversion from the `includes/Database/Models/Form.php` class and autoload those conversion files.

klhall1987:
Forms with totals or calculations seem to be breaking the forms after conversion. 

If a form has either a total or calc field then and is converted, then the form can not be edited in 3.0 and I'm seeing some errors in the PHP Error log(which can be found in the screenshot below.) 

Screenshot:
screen shot 2016-08-11 at 1 13 38 pm

Replication Steps(In 2.9.x) 
1. Create a new form. 
2. Add a Calculation field. 
3. Save form.
4. Upgrade to 3.0 and try to edit the form.

kjohnson:
@klhall1987 I removed the locking feature.

klhall1987:
It looks like any time a calculation is converted the calculation is adding a zero at the beginning of each calculation equation in 3.0. I've detailed what this looks like in a screenshot below. 

Screenshot: 
screen shot 2016-08-12 at 1 45 09 pm

Replication steps(starting in 2.9.x):
1. Create a new form. 
2. Add a number field and a total field. 
3. Set the number field to be included in the auto total.
4. Upgrade to 3.0 
5. Edit the form that was converted and check the calculation variable that was created. It should include a zero at the beginning of the equation.

klhall1987:
Forms with calculations and conditional logic are displaying "This form is not available", after conversion. If any fields on the form are moved then, the form displays correctly again. I've detailed the behavior in a screencast below. I'm not seeing a JS error in the console, but I am getting a PHP error in the PHP error log. 

Screencast:
https://goo.gl/nJDs3Z

Screenshot:
PHP error log:
screen shot 2016-08-12 at 2 20 35 pm


Replication Steps(starting in 2.9.x):
1. Create a new form.
2. Add a number, total, and textbox field. 
3. Set the number field to be included in the auto-total.
4. Make the text box show/hide based on the value of the number field. 
5. Save form. 
6. Upgrade to 3.0. 
7. Preview the form.

klhall1987:
@kjohnson 

If a form has multiple calculations and a conditional tied to one of the values of a calc field in 2.9.x, then after conversion to 3.0 the conditional will no longer fire on the front end. Everything converts correctly on the back end, but nothing is firing on the front end. I'm not seeing PHP errors besides the one that was mention in the last comment upon conversion and I'm not seeing any JS errors. I've attached a 2.9.x form export for you to test with. 

Form Export:
[SuperCalcTest---12-08-2016-.nff.zip](https://github.com/wpninjas/ninja-forms/files/416032/SuperCalcTest---12-08-2016-.nff.zip)

Replication Steps:
1. Import the form in the 2.9.x codebase. 
2. Upgrade from 2.9.x to 3.0.
3. You'll need to edit the form and a field in order to get the front end display to work. 
4. Preview the form try get the conditionals to fire on the front end.

klhall1987:
@kjohnson 

Just an update, the same behavior mentioned in the previous comment, happens with nested forms as well. I've added another Form export to test with. 

Form Export: 
[CalcTest2ElectricBoogaloo---12-08-2016-.nff.zip](https://github.com/wpninjas/ninja-forms/files/416076/CalcTest2ElectricBoogaloo---12-08-2016-.nff.zip)

kjohnson:
@klhall1987 if the first operation in an Operations Calculation uses the division operation, then PHP throws a "divide by zero" notice and the JS returns a `null` value (in v2.9.x).

kjohnson:
@klhall1987 the leading zero has been removed. The operations calculations will still processes as if there were a leading zero. This is to be expected as an advanced calculation.

kjohnson:
@klhall1987 The "This form is not available" message is a residual of the form locking feature from prior to converting calculations.

kjohnson:
@klhall1987 added a specific when type, which should fix the bug with conditions based on calculation fields.

klhall1987:
Tested and ready for merge.

PR #1602, opened 2016-08-11
Wrap mask JS in a document ready call

kjohnson:
While I am not able to replicate this issue locally, we have had similar reports of JS breaking in a similar way.


kjohnson:
Customer confirmed that this change resolves their issue.

PR #1607, opened 2016-08-12
Added a default fieldContentsData filter to both the builder and fron…

kstover:
…tend. This default filter ensures that fieldContentsData is populated with an array of field keys if no other filters are present.


kstover:
This PR changes how we handle saving and displaying forms.

@klhall1987 To test, you'll want to build forms and ensure that they are displayed properly on the front-end.

PR #1608, opened 2016-08-12
Form content

kstover:
This PR changes a few things:

1. Changes references from `fieldContents` to `formContent`. This name change is meant to convey what the variable actually represents. There are radio channels, settings, etc. that are effected.
2. The template that is used to render forms, `#nf-tmpl-form-layout`, now uses `.nf-form-content` rather than `.nf-fields`, which is more accurate.
3. Core has a system for adding filters that modify content views and loaded data, both in the builder and on the front-end. This PR moves default functionality to a filter registered with a priority of 10, rather than having the default view and data loading hard-coded in an if/else. This will give us greater flexibility, as add-ons or other code could re-use core defaults.


kstover:
This changes how we build and render forms pretty dramatically. We need to thoroughly test to make sure that neither is negatively effected.

klhall1987:
Forms are not being hidden or cleared upon submission. Everything is looks to be submitted correctly, the form just doesn't get hidden or cleared after submission. I've detailed the behavior in the screencast below. 

Screencast: 
https://goo.gl/6ZMahV

Replication steps:
1. Create a new form.
2. Add a first name, last name, and submit field. 
3. Go to emails and actions tab.
4. Add a success message, if one isn't already there. 
5. Add some text to the success message. 
6. Go to the advanced tab. 
7. Open the display settings menu, and ensure `CLEAR SUCCESSFULLY COMPLETED FORM?` and `HIDE SUCCESSFULLY COMPLETED FORM?` are turned on. 
8. Preview the form and submit.

klhall1987:
Everything seems to be working correctly expect the issue mentioned above. Thus far I have tested.

- Converting basic forms using the upgrade method. 
- Basic calculations 
- all core action types.
- basic and nested conditionals.

klhall1987:
Tested and ready for merge.

PR #1618, opened 2016-08-18
Adding a region on each side of the main content for gutters

kstover:


kstover:
This only affects the builder. @klhall1987, just test building with and without Layout & Styles active.

klhall1987:
Tested and working correctly, passing to @kjohnson for merge.

Issue #1640, opened 2016-08-25
NF_AddonChecker method forcing deactivation of NF 3 add-on


ramiabraham:
We're working on supporting NF 3 in one of our AffiliateWP add-ons, and noticed that the `NF_AddonChecker::check_plugins()` method is now forcing deactivation of this add-on.

In [this method](https://github.com/wpninjas/ninja-forms/blob/master/lib/NF_AddonChecker.php), it appears that _any_ addon will be deactivated, since the `NF_AddonChecker` class is called globally in the main Ninja Forms plugin file when NF 3 or greater is found:

https://github.com/wpninjas/ninja-forms/blob/master/ninja-forms.php#L44

```php
public function check_plugins()
    {
        if ( ! function_exists( 'get_plugins' ) ) {

            require_once ABSPATH . 'wp-admin/includes/plugin.php';

        }

        $plugins = get_plugins();

        foreach( $plugins as $plugin => $data ){

            if( 'ninja-forms/ninja-forms.php' != $plugin && 0 === strpos( $plugin, 'ninja-forms' ) && version_compare( $data[ 'Version' ], '3', '<' ) ){

                if( ! is_plugin_active( $plugin ) ) continue;

                deactivate_plugins($plugin);

                wp_redirect( admin_url( 'plugins.php?nf-deactivated=' . $plugin ) ); exit;
            }
        }
    }
```

Is this file planned to be removed or updated to allow for Ninja Forms extensions to be created?

Is there anything I can do to assist in resolving this?

Worth noting that merely commenting out the `new NF_AddonChecker` instance [here](https://github.com/wpninjas/ninja-forms/blob/master/lib/NF_AddonChecker.php#L49) completely resolves the issue.

Thanks!

kjohnson:
@ramiabraham for compatibility reasons, with the Ninja Forms THREE codebase active we disable an plugins prefixed with `ninja-forms-` if the plugin version is less than 3. The WordPress.org repository does not allow for trademarks to prefix plugins names, so we haven't run into any false positives yet.

What is the add-on of yours that is being deactivated? Is it prefixed with `ninja-forms`?

ramiabraham:
Thanks @kjohnson. The plugin slug is `affiliatewp-affiliate-forms-ninja-forms`.

ramiabraham:
Added a small pr on this @kjohnson :)

kstover:
I think that this issue is closed by merging the PR?

[...]

PR #1642, opened 2016-08-26
Use a more targeted string comparison to check for NF add-on plugins. #1640.

ramiabraham:
Switches the plugin slug check present in `NF_AddonChecker::check_plugins()` to a string comparison of the beginning of the plugin slug, checking only for a prefix `ninja-forms-`, instead of a substr of `ninja-forms` anywhere within the plugin slug.

#1640.

### DIFF
--- a/lib/NF_AddonChecker.php
+++ b/lib/NF_AddonChecker.php
@@ -20,7 +20,7 @@ final class NF_AddonChecker
 
         foreach( $plugins as $plugin => $data ){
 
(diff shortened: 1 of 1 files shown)


ramiabraham:
Ninja Forms 3, per your announcements, is set to be launched tomorrow, September fifth. As of now, without this or an equivalent adjustment, numerous add-ons will not be compatible with Ninja Forms 3.0, including AffiliateWP's _Affiliate Forms for Ninja Forms_.

Is there anything our team can do to help?

Thanks

jameslaws:
I believe @kjohnson will be merging this in tonight.

PR #1658, opened 2016-08-31
Add new merge tags

kjohnson:


klhall1987:
I am unable to use the merge tags. They are return empty. I think it has something to do with the global $post variable that is being used. Passing back to @kjohnson to look over.

kstover:
@kjohnson This needs to be looked over.

kstover:
Works for me upon further testing.

PR #1663, opened 2016-09-02
Wrap column classes in the .nf-form-cont 

### LINKED ISSUE #1577: Column classes being trumped for @media queries
Just noticed that Genesis theme column class elements are not obeying their @media queries because they are being trumped by CSS located in Ninja Forms' `/display-structure.css` file. So instead of columns being 100% width @media 1024px they are staying in their columns looking quite squished. (Pinging @jameslaws because @kjohnson said so.)

klhall1987:
so that it doesn't override use cases outside of that context. Fixes #1577 I think. :)


kstover:
@jameslaws Is this still applicable?

jameslaws:
I thought it was already in core. If it isn't, it needs to be.

jameslaws:
3.0.3 would be ideal.

PR #1678, opened 2016-09-05
Port or deprecate old functions and hooks.

kjohnson:
Added a "Deprecated Notice" for deprecated functions and hooks.

## Example

screenshot 2016-09-05 01 54 25


kjohnson:
Did a little spot checking, doesn't seem to break anything.

Issue #1720, opened 2016-09-08
Make 'Edit' link on form preview page actually edit the form not the preview page


polevaultweb:
![screen shot 2016-09-08 at 16 50 08](https://cloud.githubusercontent.com/assets/1770201/18356403/6f13d916-75e4-11e6-8c57-e1aa35bf72bc.png)

Currently the link points to http://localhost/nftest/wp/wp-admin/post.php?post=1&action=edit

kstover:
@kjohnson What do you think?

polevaultweb:
@kstover The use case for this, if I open Chrome with no tabs, start typing my site and see the auto complete options for it, including the form preview. If i click the preview - there is no easy way to get to the form edit. Small UX improvement imo - easy fix in https://github.com/wpninjas/ninja-forms/pull/1721

kjohnson:
As per conversation with @polevaultweb, the "Edit" link should just be overridden with an `href` value of `#` if the link cannot just be removed.

My problem is that if you are editing the form in another tab, preview the form, then click edit, the builder opens in yet another tab without the previous changes.

jameslaws:
Issues in this repo have become unwieldy and in many cases unreliable at identifying actual problems. In a first step to start to remdy this we are doing a clean up of older issues. 

This is issue is being closed, but if you can confirm this as an actual current bug in the 3.x codebase and the new UI, please re-open or comment and we will review it.

PR #1721, opened 2016-09-08
Override the edit post URL with the edit form URL in the form preview

polevaultweb:
Resolves https://github.com/wpninjas/ninja-forms/issues/1720


kjohnson:
Ah, brilliant.

kstover:
@kjohnson @polevaultweb Would this mean clicking the "Edit" button takes you to edit the form instead of the post?

jameslaws:
If we can do it easy enough, I think this makes sense on the preview page since it's not actually a page.

kjohnson:
@kstover yes, as opposed to editing the page that the preview leverages.
@jameslaws I agree that this would be more contextually correct.

polevaultweb:
@kjohnson `master` merged in, no conflicts

PR #1729, opened 2016-09-08
Update default input_limit_type check.

### LINKED ISSUE #1716: Character count broken in 3.0 single line text fields
image

In the above screenshot in 3.0.2, the character limit is set to "100"

It appears to be counting words instead of characters.

kjohnson:
Closes #1716.


klhall1987:
Tested, working, and passing to @kstover for merge.

PR #1735, opened 2016-09-09
Use form instance id instead of form id for display

kjohnson:
Duplicate form IDs are bypassed in the `FormCollection`, which means that the same form cannot be displayed twice on the same page.

Instead of using the form ID, I've generated an instance ID to use instead and updated the reference in the fields.


kjohnson:
Re-working to a JS display only solution.

PR #1753, opened 2016-09-11
Add notice to advanced rollback setting

### LINKED ISSUE #1747: Notify user that rollback causes data loss
We need to inform users that the rollback to 2.9 removes all 3.0 data.  This is not currently understood and has caused some users data loss since it's currently our first recommended troubleshooting step.

I would suggest this notice:

"Please note that all 3.0 data will be removed.  Please export any forms or submissions you do not want to be lost during this process."

kjohnson:
Closes #1747.


kjohnson:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;dSrSsRiB&#x2F;37-notify-user-that-rollback-causes-data-loss"> Notify user that rollback causes data loss</a>

klhall1987:
Tested, working, merging

PR #1803, opened 2016-09-15
Set the session ID on `plugins_loaded`. Closes #1649

### LINKED ISSUE #1649: Session regenerated on every page - version 2.9.56.2
This has been an issue before with the 2.9 branch. See #1236, #954, #1502. It's an issue again, with version 2.9.56.2 (not using v3).

I was able to check this with a clean install of WP with no other plugins installed and the default theme.

ericnicolaas:


kstover:
@ericnicolaas Sorry for the delay in looking at this! We're working on our development processes, including responding to PRs in a more timely manner.

@kjohnson What are the ramifications of this? Are we still using the session manager?

kjohnson:
@kstover we had made modifications to set the cookie only as needed, but I believe that we have stopped using the session manager, due to related complications.

kjohnson:
@ericnicolaas are you using the session manager for anything in particular?

ericnicolaas:
@kjohnson We use `WP_Session` in our plugin, Charitable (https://github.com/Charitable/Charitable). Since WP_Session is a common library that many plugins use, it's important to make sure it's reliable. Your implementation, where the session is being reset on every page load, means that sessions are effectively useless. Anything we store in the session is lost by the next page load. 

If you're not using WP_Session in Ninja Forms, can you get rid of it entirely?

PR #1812, opened 2016-09-17
Attributes are not escaped preventing refresh button to be displayed/functional.

mch0lic:
This affects chrome, but might as well affect other browsers too.


kstover:
Hey @mch0lic. I'm very sorry for the super-late reply! We actually took care of this a while ago by removing the localization altogether. It shouldn't have been localized with __().

PR #1821, opened 2016-09-20
Suppress HTML fields from all fields merge tags

### LINKED ISSUE #1816: Remove HTML fields from table output in all fields merge tag
HTML layout field content should not appear in the "all fields" merge tag but should still be available as an individual merge tag.

kjohnson:
Closes #1816.


klhall1987:
Tested, Working correctly, merging.

PR #1828, opened 2016-09-21
Defer render of reCaptcha

### LINKED ISSUE #1805: reCaptchas showing intermittently
reCaptchas showing intermittently in 3.0
When this happens there are no errors in the console.

Here are the current cases for users having this issue:
https://wpninjas.desk.com/web/agent/search/cases?query=labels:recaptcha%20status:open

kjohnson:
Closes #1805.


klhall1987:
I sent this to a customer to test and here are the issues he was seeing still seeing. I did not confirm on my own yet. I'll test this with browser stack a bit later. 

Chrome -> Still intermittent on page load
Internet Explorer -> Does not work at all now on page load, even after multiple refreshes
Edge -> Does not work at all now on page load, even after multiple refreshes
Firefox -> Appears to work on every page load

PR #1840, opened 2016-09-22
Closes Issue#1832

kstover:
Fixes a bug with hidden fields and calculations.

Calculations should be tested fully before merging. (Test new calcs using hidden fields, but also without them)


klhall1987:
Tested, working correctly, and merging.

PR #1905, opened 2016-10-03
Fields that have a visible attribute of false should now not output a…

### LINKED ISSUE #1877: Fields with `visible` set to `false` output wrapper HTML
When a field has a `visible` attribute of `false` the field element is hidden, but the wrapping HTML is still displayed.

kstover:
…ny content. Closes #1877


klhall1987:

PR #1921, opened 2016-10-06
Adding a radio message that is triggered when a pikaday instance is initialized

### LINKED ISSUE #1920: Radio message on pikaday init
After we init a pikaday instance, we should fire a radio message so that the instance can be customised.

kstover:
Closes #1920.


klhall1987:

PR #1926, opened 2016-10-06
Ensuring that dashicons are enqueued on the front-end. Closes #1925.

### LINKED ISSUE #1925: RTE buttons don't show for non logged-in users
If a user isn't logged-in, they won't see any RTE buttons output by the textarea field display setting.

kstover:


klhall1987:

PR #1928, opened 2016-10-06
Adding an empty function for ninja_forms_get_form_by_id() so that sit…

### LINKED ISSUE #1927: Deprecate ninja_forms_get_form_by_id() function
Some users are seeing Fatal PHP errors because other plugins attempt to use ninja_forms_get_form_by_id(), but it hasn't been deprecated. To be fair, this function was deprecated by version 2.2. We should include a function so that the fatal error doesn't occur.

kstover:
…es that utilize the older function do not throw a fatal error. Closes #1927.

### DIFF
--- a/includes/deprecated.php
+++ b/includes/deprecated.php
@@ -289,3 +289,7 @@ function _nf_removed_hooks() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


klhall1987:
tested, working, merging into develop branch.

PR #1936, opened 2016-10-12
Fix object ID comparison for loading field settings from cache

### LINKED ISSUE #1935: Weird UI js/data issues
Feel free to rename the title not sure how to classify this issue ..

We've started to use Ninja Forms for client installations, our most recent install uses nf3 (v3.0.8). But the UI seems to be a bit buggy. I've used it in Safari as well as Firefox (local install and WPEngine install) and I get the same behavior.

I create a simple form, both from the default "contact us" form and from scratch, after saving or after update, certain fields aren't recognized, the screen looks like the attached screenshot. The fields cant be read and display "false", clicking to edit the field yields a message: Field type "" not found."

screen shot 2016-10-11 at 4 02 31 pm

There is definitely some weirdness going on, which is quite troubling.

**EDIT:**

I tried uninstalling, I check "remove all data" before uninstall. I then reinstalled, but all my form data was still available (all fields were "false" as pictured in screenshot)

Tested using Chrome, similar behavior, I add a field and publish, exit, and reenter the form and the field comes up as "false".

Uninstalled plugin .. went into DB and deleted all nf3, nf and ninja_forms tables .. reinstalled, went into the default "contact me" form, saw an old field I added and double submit field (weird).

Renaming fields and changing field values seem not to work either, it always reverts back to the old values when i exit and reenter the form.

kjohnson:
Closes #1935 .

### DIFF
--- a/includes/Abstracts/Model.php
+++ b/includes/Abstracts/Model.php
@@ -227,8 +227,8 @@ class NF_Abstracts_Model
 
                 if (isset($form_cache[ 'fields' ])) {
 
-                    foreach ($form_cache[ 'fields' ] as $object_id => $object) {
-                        if ($this->_id != $object_id) continue;
+                    foreach ($form_cache[ 'fields' ] as $object) {
+                        if ($this->_id != $object[ 'id' ]) continue;
 
                         $this->update_settings($object['settings']);
                         break;


kjohnson:
Tested, successfully, with [case#24600](https://wpninjas.desk.com/web/agent/case/24600).

Before the update I was able to create a form, publish, and refresh the page before the background processing completed. This caused the builder to load before field settings were saved, such as field type, which would then cause empty fields to load with unknown types. A publish action from this loaded state could then save these undefined fields as new fields. This issue is exacerbated by slower server response times.

After the update the form re-loaded from the saved form cache and does not cause the issue stated above.

Issue #1950, opened 2016-10-14
Form cache doesn't work for new fields


kstover:
If you add a new field to a form (including initial creation), then publish, the cache stores the temporary ID. When the field later looks for itself in the cache, it has a real ID and can't find itself.

PR #1952, opened 2016-10-14
Updating temporary field IDs in the form cache with their real IDs. C…

kstover:
…loses #1950.


klhall1987:
@kjohnson It looks like processing data is still being pull from the database.

kjohnson:
@klhall can you elaborate?

jameslaws:
Let me try. The Form is still using the field data from the database for actions on processing instead of from the cache.

klhall1987:
@kjohnson said the issue mentioned above is being dealt with in a separate branch(https://github.com/wpninjas/ninja-forms/tree/issue%231950-2). So this branch is tested, working, and ready to merge.

PR #1966, opened 2016-10-17
Adding an option to set the string "checked" or "unchecked" for check…

### LINKED ISSUE #1946: Checkbox doesn't pass data to email actions, success message, etc
Single checkbox impacted, does not appear to impact multi select lists.

kstover:
…box merge tags. Fixing a bug that caused checkboxes to return empty strings when used in merge tags. Closes #1946


kstover:
Better handled in #1970

Issue #1993, opened 2016-10-21
Add Custom Currency Symbol Setting


klhall1987:
We need to have a setting to allow users to set a custom currency symbol.

klhall1987:
Requested again by [Gregory](https://wpninjas.kustomerapp.com/app/customers/5878ce392498d81000631580/event/5878ce39d6f7121100e68c84)

jameslaws:
Issues in this repo have become unwieldy and in many cases unreliable at identifying actual problems. In a first step to start to remdy this we are doing a clean up of older issues. 

This is issue is being closed, but if you can confirm this as an actual current bug in the 3.x codebase and the new UI, please re-open or comment and we will review it.

PR #2000, opened 2016-10-24
Add Custom Currency Symbol Setting

paqu83:
My proposition of solving this #1993


kstover:
@paqu83 Hey! Sorry for the massive delay in getting to this. We have some built-in methods of adding settings with dependencies, so you don't have to do all the JS lifting manually. I'm going to close this PR, but I'll ping some devs to get this moving in the right direction.

@klhall1987 Could you get with @kjohnson and work out a cleaner way of adding the custom currency symbol?

Issue #2001, opened 2016-10-25
Submission count does not increment


kjohnson:

jameslaws:
@kjohnson You told me yesterday that this was patched and merged, but it obviously was not...

This still isn't working. Just tested with an up to date Develop branch and all submissions come through as zero: https://db.tt/UiK9bG1I

jameslaws:
I rolled back to 2.9 to check a few things and I had no forms. I created a new form with a single textbox and submitted once. When I view the submissions I see the 1 I submitted and what appears to be 5 duplicates dated 7 days earlier: https://db.tt/aPAmaQHl

So there is weirdness in 2.9 codebase as well.

jameslaws:
Issues in this repo have become unwieldy and in many cases unreliable at identifying actual problems. In a first step to start to remdy this we are doing a clean up of older issues. 

This is issue is being closed, but if you can confirm this as an actual current bug in the 3.x codebase and the new UI, please re-open or comment and we will review it.

boluda:
Hi there! The submission count field doesn't increment either.

![image](https://user-images.githubusercontent.com/4599915/39099047-f5cf6c34-4673-11e8-88d5-e48d69fd6f3f.png)

It always returns 0 :(

robertogcalle:
Hi, guys.
This issue has to do with the query used inside the function that returns the number of submissions.
All you have to do is change the code of the function get_sub_count() in the Form.php file located at ninja-forms->includes->MergeTags.
```
public function get_sub_count()
    {
		global $wpdb;

		// Query the database for the total amount of submissions for a form.
/*    	$query = "SELECT DISTINCT( COUNT( wpp.id ) ) AS sub_count FROM `" . // That DISCTINCT is the problem
			$wpdb->prefix . "posts` wpp JOIN 
`" . $wpdb->prefix . "postmeta` 
			wpm ON wpp.id = wpm.post_id
			WHERE wpm.meta_key = '_form_id' AND wpm.meta_value = %s";
*/
        $query = "SELECT COUNT( wpp.id ) AS sub_count FROM `" .
            $wpdb->prefix . "posts` wpp JOIN 
            `" . $wpdb->prefix . "postmeta` 
            wpm ON wpp.id = wpm.post_id
            WHERE wpm.meta_key = '_form_id' AND wpm.meta_value = %s";
    	$count = $wpdb->get_results( $wpdb->prepare( $query, $this->form_id ) );

[...]

PR #2008, opened 2016-10-25
bypass cache for submission sequence.

kjohnson:
The submission sequence number was being pulled from the cache, which wasn't being updated.

This will need to be modified again when we update the Models and Cache.


kjohnson:
Closes #2001.

PR #2015, opened 2016-10-26
Added ID attribute to the field container

### LINKED ISSUE #2014: Add field ID to field container
Currently it is not possible to target the field container specifically at all.

https://wpninjas.desk.com/web/agent/case/25447

Is there a way to do with with the class name as well?

jameslaws:
Closes #2014


klhall1987:
Tested and working. Merging into develop branch.

PR #2039, opened 2016-11-02
Add get_fields fallback if form cache does not exist.

kjohnson:


jameslaws:
After deleting cache I resubmitted the form and when I reload the submission screen to make sure it everything is there I see this: https://db.tt/sSLQeDZU

Warning: Invalid argument supplied for foreach() in /Users/james/Documents/dev/WordPress/wp-content/plugins/local-three/includes/Admin/CPT/Submission.php on line 124

PR #2049, opened 2016-11-03
Un-translate programatic value.

kjohnson:


klhall1987:
Fix tested and working correctly.

PR #2081, opened 2016-11-09
Halt submission if field validation throws an error.

kjohnson:
If field validation throws an error, halt the submission before fields or actions can process.


kjohnson:
This also addresses the issue of the reCaptcha failure not halting form submission.

PR #2086, opened 2016-11-09
Add compatibility for locale number format in product fields.

kjohnson:


kstover:
@kjohnson This has the `in progress` label. Is it ready to test?

kjohnson:
@kstover while this seems to work locally, I'm getting reports from a user that does not work for them. I'm probably going to re-evaluate and re-submit a new PR.

Issue #1298, opened 2016-04-13
Allow multiple recaptcha fields on the same page


klhall1987:
A user has requested the ability to add multiple recaptcha fields on a page. Currently if a page has multiple recaptcha fields on a page, only one recaptcha field will show and a console error is thrown that states, "Uncaught Error: ReCAPTCHA placeholder element must be empty".  The user( Revxx14 ) sent us a gist showing us how they modified the ninja-forms-recaptcha.php in 2.9.x to solve this issue. 

Revxx14's GIst:
https://gist.github.com/revxx14/544a72290e042293164d169b10a92318

hitittome873:
Found a fix to this issue by adding the explicit render flag on the recaptcha api.js file call from google.

<script type="text/javascript" src="https://www.google.com/recaptcha/api.js?**render=explicit**&hl=<?php echo $lang; ?>"> </script>

hitittome873:
Anyone able to implement a change on this? I've managed to get it working by using a random id per captcha container as a selector as well as the explicit render option to load the captcha script.

Here's my current display function:
`function ninja_forms_field_recaptcha_display( $field_id, $data, $form_id = '' ) {
	$settings = get_option( "ninja_forms_settings" );
	$lang = $settings['recaptcha_lang'];
	$siteKey = $settings['recaptcha_site_key'];
	$field_class = ninja_forms_get_field_class( $field_id, $form_id );
	$rand = wp_rand(0, 99999);
	wp_enqueue_script(
		'g-recaptcha',
		'https://www.google.com/recaptcha/api.js?onload=nf_grecaptcha_explicit_render&render=explicit&hl='.$lang, ['nf-grecaptcha-explicit-render'] );
	if ( !empty( $siteKey ) ) { ?>
		<input id="ninja_forms_field_<?php echo $rand;?>" name="ninja_forms_field_<?php echo $field_id;?>" type="hidden" class="<?php echo $field_class;?>" value="" rel="<?php echo $field_id;?>" />
		<div id="g-recaptcha-<?php echo $rand; ?>" class="g-recaptcha" data-field-id="<?= $rand; ?>" data-callback="nf_recaptcha_set_field_value" data-sitekey="<?php echo $siteKey; ?>"></div>
		<script type="text/javascript">
            function nf_recaptcha_set_field_value(inpval){
            	jQuery("#ninja_forms_field_<?php echo $rand;?>").val(inpval)
            }
            </script>
		<?php
	}
}`

[...]

PR #2096, opened 2016-11-11
Add random id and explicit rendering function for captchas #1298

mycarrysun:
Added new js function for explicit rendering called ninja_forms_grecaptcha_explicit_render() to allow explicit rendering of captchas on a single page. I have only implemented this for plugins below version 3.0 (in the deprecated directory). Also added a unique ID to select each captcha input field on success (for nf_recaptcha_set_field_value()).

* Enqueue google recaptcha script with the ninja-forms-display dependency
* Added function for explicit rendering of captcha fields
* Changed ID of recaptcha hidden field to have a random number instead of the field ID (we need unique field ID's for explicit render)
* Removed duplicate array key 'sidebar'

### DIFF
--- a/deprecated/includes/fields/recaptcha.php
+++ b/deprecated/includes/fields/recaptcha.php
@@ -17,7 +17,6 @@ function ninja_forms_register_field_recaptcha() {
 		'edit_custom_class' => false,
 		'edit_help' => false,
 		'edit_meta' => false,
-		'sidebar' => 'template_fields',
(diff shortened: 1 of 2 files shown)


kstover:
Hey @mycarrysun. I'm a little tired; I merged the PR, but missed that it was against the master branch. I can't re-open it or change the branch now, but could you re-submit on the `develop` branch?

PR #2125, opened 2016-11-18
Field keys now use type_{timestamp} rather than type_{#} as default. …

### LINKED ISSUE #2124: Duplication Bug
When you duplicate a form field, then add a new field of the same type, the field key of the newly added field will be the same as the "copy" field that was just created via duplication.

kstover:
…Closes #2124.


klhall1987:
@kstover tested on users sites. Merging into develop now.

PR #2129, opened 2016-11-21
Settings no longer re-render by default. Setting types can still spec…

### LINKED ISSUE #2108: Settings in the drawer do not need to re-render by default on change
Textboxes and other settings in Drawers automatically re-render when they are changed. This probably shouldn't be the default behaviour, since it can mess up stuff like acceptance testing.

kstover:
…ify that they would like to re-render. Also added a class to advanced boxes. 

To Test: 

- Edit as many settings as you can. Make sure that the setting drawer renders all the settings properly.

Closes #2108.


klhall1987:
Tested, everything seems to be working correctly. Resolving merge conflicts and merging into develop.

PR #2179, opened 2016-12-06
Issue/1298

mycarrysun:
Added new js function for explicit rendering called ninja_forms_grecaptcha_explicit_render() to allow explicit rendering of captchas on a single page. I have only implemented this for plugins below version 3.0 (in the deprecated directory). Also added a unique ID to select each captcha input field on success (for nf_recaptcha_set_field_value()).

- Enqueue google recaptcha script with the ninja-forms-display dependency
- Added function for explicit rendering of captcha fields
- Changed ID of recaptcha hidden field to have a random number instead of the field ID (we need unique field ID's for explicit render)
- Removed duplicate array key 'sidebar'

@kstover my bad for PR'ing into the master branch! First time contributor. Let me know if this is the right way to do it.


kstover:
It looks like this breaks reCaptcha in the deprecated codebase.

The reCaptcha loads initially, but never allows you to submit the form. It just gives the "This is a required field" error. It worked in my initial test, but has since broken. I'm not sure what about it is breaking, and unfortunately I don't have time to track this down for the deprecated codebase, so I'm going to revert the changes to fix reCaptcha in 2.9.x. 

@mycarrysun, If you want to test and get back to me, I'd be happy to take another look.

mycarrysun:
@kstover It's broken for me too...I'm not sure what's causing it either. I don't have time at the moment either to debug but no worries. I will get to it sometime this month, I think I have a fix for it in another site, just need to test and make sure it works and then create a PR.

kstover:
@mycarrysun No problem. I'm about to release 3.0.24, which reverts the change. We can circle back around when you get time.

Issue #1584, opened 2016-08-09
NF_Admin_AllFormsTable might have to set after admin_init as $GLOBALS['hook_suffix'] is still not defined


yeetien:
I'm using add_action( "admin_post_{$action}" ) and encountered "Notice: Undefined index: hook_suffix in wp-admin/includes/class-wp-screen.php on line 229".

Here's the stack:

![capture](https://cloud.githubusercontent.com/assets/2802794/17507554/ced5e3f8-5e42-11e6-898b-1be4a02c01fa.PNG)

Seems like it's related to https://core.trac.wordpress.org/ticket/29933.

Possible to move `$this->table = new NF_Admin_AllFormsTable();` (.../Forms.php:31) to a action after `admin_init`?

NF version: 2.9.55.1 (running 3.0)
WP version: 4.5.3

kjohnson:
@yeetien can you provide a code example of what you are trying to do?

yeetien:
Hi @kjohnson 

Sorry, didn't explain clearly...

Below is the code of what I'm trying to do. I got "Undefined index: hook_suffix" when accessing http://localhost/wp-admin/admin-post.php?action=export. 

```
add_action('manage_posts_extra_tablenav', function($which)
{
	echo '<div class="alignleft actions">
		<a href="' . admin_url('admin-post.php?action=export') . '" class="button">Export</a>
	</div>';
});

add_action('admin_post_export', function()
{
	header('Content-Type: application/csv');
    header('Content-Disposition: attachment; filename=export.csv');
    header('Pragma: no-cache');
	
    $output = fopen('php://output', 'w');
	fputcsv( $output, array(
		'Name', 
		'ID',		
	) );
	return $output;
});
```

o1y:
This bug breaks the hook [admin_post_{$action}](https://developer.wordpress.org/reference/hooks/admin_post_action/). For testing just put a random param like `?action=foobar` to `admin-post.php` - no more code required.

Maybe you can use the action hook `current_screen` instead of `admin_init`, in [Forms.php](https://github.com/wpninjas/ninja-forms/blob/master/includes/Admin/Menus/Forms.php#L18)?

simmerer:
@o1y That seems to work for me.

simmerer:
(Sorry for all the notification churn here. Finally got my shit together.)

kstover:
Should be fixed in the latest version. Closing for now. If this is still an issue, feel free to re-open.

PR #2217, opened 2016-12-19
issue/1584

simmerer:
Unbreak the admin_post_($action} hook by using 'current_screen' instead of 'admin_init', as recommended by @o1y in https://github.com/wpninjas/ninja-forms/issues/1584#issuecomment-264245151


kjohnson:
@simmerer this looks good to me. I'm going to pass this along for testing.

PR #2224, opened 2016-12-19
Fixing a bug with updating plugins running the deprecated codebase. C…

kstover:
…loses #2323.


klhall1987:
Fix tested and working properly.

PR #2245, opened 2016-12-22
locale decimal update (issue#2244)

### LINKED ISSUE #2244: decimal separator disapear
I use 3.0.20 with product fields in euros € on FR website
When I import my FR form on my EN website,
decimal separator is lost :
€12,00 becomes €1,200
€1,00 becomes €100

rlagarde:
I use this tofloat function to get a clean float from any locale string number.
it uses floatval, I then use sprintf instead of number_format.

This solved first issue I had with decimals not displaying at all in a previous nf version

I just changed it here for product price. didn't touch any other appearance of number_format.

-- edit -- 
Closes #2244.


kjohnson:
@rlagarde I thought that we had already addressed this in a recent release. Are you still seeing missing currency separators as of v3.0.20?

rlagarde:
This is happening on 3.0.20 as I described in issue#2244

kstover:
@kjohnson Could you respond?

kjohnson:
@rlagarde sorry that this has taken so long for us to review this. We are still trying to workout a system for better handling community contributions. Also, we had a Products sprint scheduled that has been delayed, so this hasn't been a high priority for the project.

I'm going through and reviewing these changes.

kjohnson:
@rlagarde looking more into #2244, the issue seems more to be with the builder side of things.

What is happening is that the Product Price setting is being stored with the comma separator, so when switching to English the comma is removed as an improperly placed thousands separator.

This issues seems to be more systemic and should be resolved with how that value is stored, otherwise #2244 would not be resolved.

For you particular issue of importing from FR to EN, I would suggest manually updating the Product Price setting of the product field.

Issue #2249, opened 2016-12-23
UNDEFINED INDEX: CALLBACK IN [...] MERGETAGS.PHP


isleshocky77:
So this looks to be an issue from the change between v2 and v3. When you select "Query String" in the "Merge Tags" pop up for a default value it inserts "{query_string_key}". This actually incorrect as it's not useful. The new way of doing it is to just use any key that will be put in the querystring. "{email}" and then http://website.com?email=email@example.org.

However, if you leave "{query_string_key}" there and attempt to use it you get the following error.  You'll have this issue if you add a new "MergeTag" subclass which does not have a call back.

There are two options to fix this.

## Fix for QueryStrings.php

```diff
Index: includes/MergeTags/QueryStrings.php
===================================================================
--- includes/MergeTags/QueryStrings.php	(revision 1559357)
+++ includes/MergeTags/QueryStrings.php	(working copy)
@@ -16,6 +16,7 @@
             '' => array(
                 'tag' => '{query_string_key}',
                 'label' => __( 'Query String', 'ninja_forms' ),
+                'callback' => null,
             ),
         );
```

## Fix for MergeTags.php

```diff
Index: includes/Abstracts/MergeTags.php
===================================================================
--- includes/Abstracts/MergeTags.php	(revision 1559357)
+++ includes/Abstracts/MergeTags.php	(working copy)
@@ -47,6 +47,8 @@
         foreach( $this->merge_tags as $merge_tag ){
             if( ! in_array( $merge_tag[ 'tag' ], $matches[0] ) ) continue;
 
+            if( ! isset($merge_tag[ 'callback' ])) continue;
+
             $replace = ( is_callable( array( $this, $merge_tag[ 'callback' ] ) ) ) ? $this->{$merge_tag[ 'callback' ]}() : '';
 
             $subject = str_replace( $merge_tag[ 'tag' ], $replace, $subject );
```

## The Error and Trace

```
( ! ) NOTICE: UNDEFINED INDEX: CALLBACK IN /SRV/WEBSITE/PROJECT/SITE/WEB/APP/PLUGINS/NINJA-FORMS/INCLUDES/ABSTRACTS/MERGETAGS.PHP ON LINE 50
CALL STACK
#	TIME	MEMORY	FUNCTION	LOCATION
1	0.0013	235104	{main}( )	.../index.php:0
2	0.0355	236856	require( '/srv/website/project/site/web/wp/wp-blog-header.php' )	.../index.php:5
[...]

PR #2250, opened 2016-12-23
Undefined index "callback" in NF_Abstracts_MergeTags::replace for QueryStrings (Issue #2249)

isleshocky77:
This resolves issues in #2249 by checking for a defined callback before attempting to use it.


isleshocky77:
@kstover Which tag did this make it into?

Issue #2258, opened 2016-12-29
"Export" post row action for nf_sub post type displays on all custom post types (posts too, I think)


thomasplevy:
see pr #2259 for possible fix

PR #2259, opened 2016-12-29
Add check for submission export post row action

thomasplevy:
fix for issue #2258


thomasplevy:
@kjohnson that's prettier, nice call, see latest commit

PR #2273, opened 2017-01-03
Add filter on data after action has been processed.

polevaultweb:
This is so field processing can hook in to the save action (the first instance the submission ID is available).

Table Editor use case - the CSV is created using the [field label]-[sub_id].csv, but the CSV can't be created on field `process`.

Adding more data to the save sub hooks won't help, as TE needs to enrich the values/field object passed around to other actions, and the save sub hooks are actions not filters.


polevaultweb:
I have discussed with @kjohnson but a further discussion needs to happen

polevaultweb:
I'm close to getting Table Editor ready for release, could we fit this into 3.0.21 @kstover @kjohnson?

PR #2283, opened 2017-01-05
Add field mergetag add method

polevaultweb:
So field mergetags can be added and updated by other code.

As discussed by @kjohnson. Needed for Table Editor addon to update field value after save submission action.


polevaultweb:
I'm close to getting Table Editor ready for release, could we fit this into `3.0.21` @kstover @kjohnson?

PR #2295, opened 2017-01-09
Powered by Ninja Forms

kjohnson:
Adds a new `ninja` field type for displaying a "Powered by Ninja Forms" message.

poweredbyninjaforms


kstover:
I think we need to discuss this as a team. Not sure that the .org repo would look too kindly on this kind of advertising.

Issue #2308, opened 2017-01-13
Add a password type for plugin settings


kjohnson:
For account passwords and secret keys.

See: https://github.com/wpninjas/ninja-forms/blob/master/includes/Templates/admin-menu-settings.html.php#L49

PR #2309, opened 2017-01-13
Add password setting type

stuart-sequeira:
For issue 2308

Creates a new type of setting, called 'password'

Sets the input type to password so that it doesn't display the value on the settings page.  Value is still stored as plain text


kjohnson:
Closes #2308.

PR #2318, opened 2017-01-17
Update field sort ordering

### LINKED ISSUE #2236: Duplicated forms do not retain proper field order
When duplicating a form with multiple fields, the duplicates do not retain the order of fields from the original; i.e. the field order changes from duplicate to duplicate. 

First reported here: https://wpninjas.kustomerapp.com/app/customers/583867da4bcc3c1100baefc2/event/585826bff06ae310000496f8

kjohnson:
Closes #2236.


kjohnson:
The localized field order is now right, but the JS is still sorting as strings.

kjohnson:
@rlagarde try it now.

rlagarde:
@kjohnson thanks. It's working

PR #2319, opened 2017-01-18
Update field sort ordering on form edit page

### LINKED ISSUE #2236: Duplicated forms do not retain proper field order

rlagarde:
Closes #2236.

- Cast order to integer
- Sort array before storing to preloadedFormData
- Add array_sort function


kjohnson:
@rlagarde this is too closely coupled with form data localization, where as it needs to be applied where it covers localization, export, etc.

rlagarde:
@kjohnson 
ok as you wish.
I just wanted to share this fix because it worked for me and issue #2236 has been open for a month.
Fields are now sorted in correct order with duplicated forms, imported/exported forms, template loaded forms...

I'm not sure if I get when this won't work.

PR #2323, opened 2017-01-18
Update MergeTags.php

apermo:
Enhanced the callback capability, by allowing to call outside functions, instead of only functions from within the scope.

Explanation: Without this addition there is no way to add own MergeTags via 'ninja_forms_merge_tags_post' inside the /includes/Config/MergeTagsPost.php as this only applied to function within a final class.

### DIFF
--- a/includes/Abstracts/MergeTags.php
+++ b/includes/Abstracts/MergeTags.php
@@ -49,8 +49,14 @@ abstract class NF_Abstracts_MergeTags
 
             if( ! isset($merge_tag[ 'callback' ])) continue;
 
-            $replace = ( is_callable( array( $this, $merge_tag[ 'callback' ] ) ) ) ? $this->{$merge_tag[ 'callback' ]}() : '';
-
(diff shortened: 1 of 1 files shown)


kjohnson:
For testing purposes:

```php
<?php

/*
 * Plugin Name: Foo
 */

add_filter( 'ninja_forms_merge_tags_post', function( $tags ){

    $tags[ 'foo' ] = array(
        'id' => 'foo',
        'tag' => '{post:foo}',
        'label' => __( 'Foo', 'ninja_forms' ),
        'callback' => 'foo_merge_tag'
    );
    return $tags;
});

function foo_merge_tag(){
    return 'bar';
}
```

kstover:
@kjohnson What's the status of this PR?

kjohnson:
@kstover I'm working through the Merge Tag refactor and I am looking at including this PR on that release.

kjohnson:
@apermo hoping to get a closer look this weekend. This goes along with the merge tag refactor and seems like a good fit.

kjohnson:
@kstover Looks good to me!

Issue #2261, opened 2016-12-29
Unable to add favorite list fields to the form


jmcelhaney:
Reproduction steps:
Create a drop-down list and save it as a favorite
Go to another form and click on the favorite field to add it to the form (the field is not added)
Drag the field into the form and try to publish (a javascript error will appear in the console: Uncaught TypeError: Cannot read property 'get' of undefined coming from /ninja-forms/assets/js/min/builder.js)

Also when Layout & Styles is active and you try to drag the field to the form, the other fields on the form disappear

### Environment ###
Home URL:                http://localhost/wordpress3
Site URL:                http://localhost/wordpress3
Ninja Forms Version:     3.0.20
WP Version:              4.7
WP Multisite Enabled:    No
Web Server Info:         Apache/2.2.29 (Unix) mod_wsgi/3.5 Python/2.7.10 PHP/7.0.8 mod_ssl/2.2.29 OpenSSL/0.9.8zh DAV/2 mod_fastcgi/2.4.6 mod_perl/2.0.9 Perl/v5.22.0
PHP Version:             7.0.8
MySQL Version:           5.5.42
PHP Locale:              negative_sign:
WP Memory Limit:         40M
WP Debug Mode:           Yes
WP Language:             Default
WP Max Upload Size:      32 MB
PHP Post Max Size:       32M
Max Input Nesting Level: 64
PHP Time Limit:          30
PHP Max Input Vars:      1000
SUHOSIN Installed:       No
Server IP Address:       ::1
Host Name:               localhost
SMTP:                    localhost
smtp_port:               25
Default Timezone:        America/New_York

### Plugins ###
Installed Plugins:       Ninja Forms - Layout & Styles by The WP Ninjas version 3.0.12
                         Ninja Forms by The WP Ninjas version 3.0.20

PR #2332, opened 2017-01-18
Fixing a bug that caused saved fields to fail to add properly to the …

### LINKED ISSUE #1916: Saved/favorited "Select" fields cannot be used in the builder
fieldItem.js:68 Uncaught TypeError: Cannot read property 'get' of undefined

This error happens when trying to add a "Select" saved field into the builder.

When added, a glitched saved field with no controls appears in the form with the error above.  

Reproduction steps:
1. Add a "Select" field
2. Make it a "Saved Field" (you can make changes or leave settings as default values)
3. Add the field to the form by dragging it into the builder (note: click to add does not work at all)
4. Only the select field adds to form, sans icons/styling/ability to open configuration drawer.  JavaScript error appears on page.

NOTE:  With LAyout and Styles/MPF activated, the above steps happen but all other fields are visually removed from the form builder as well.  The video below demonstrates this behavior ONLY in vanilla Ninja Forms.

https://goo.gl/YUWzr8

kstover:
…form. Closes #1916.

There are also PRs in for MP and Layout & Styles.


kjohnson:
Closes #2261.

PR #1898, opened 2016-10-03
Add Metaboxes to Ninja Forms Instance

cloudratha:
To improve accessibility to the AppendAForm Metabox, and any future Metaboxes, a public property has been provided.

PR #2335, opened 2017-01-19
Stores the class instance responsible for displaying the Append A Ninja Form metabox.

jlambe:
For the context if this PR. I have a project where I assign a page template. The page template has a 2 columns layout. The left column is filled with the `post_content` of the page and the right column is filled with a custom contact/registration form.

For the user, I want him to be able to control which form he would like to add to his page and thought why not use the `form ID` defined in the `Append Ninja Form` metabox. Inside my template, in the right column HTML code, I can simply call the ninja form function and pass the user defined form ID.

It works perfectly and brings control to the user without adding complexity to the admin. But the issue is that the class responsible for the metabox also append a shortcode automatically at the end of the `post_content`. So I needed to call the `remove_filter` function, pass it the original instance that registers the filter and its method in order to be able to de-activate the append behavior for my specific template.

The PR here, only stores the `NF_Admin_Metaboxes_AppendAForm` instance into plugin Ninja Form main instance. So if a want to remove the default behavior of the append form, I can do this:
```php
/*
 * Remove the ninja form appended form from the page content.
 */
$nf = Ninja_Forms();
remove_filter('the_content', [$nf->metabox, 'append_form']);
```

This provides a clean and easy syntax to remove the default behavior of the Append A Ninja Form feature and re-use it to display the form anywhere from a custom template.

Let me know what you think of this! I'm sure it will be useful for many developers/users.


jlambe:
Hi @kstover,
do you have any merged date regarding this PR? We do have a project that rely on this change and we would love to stop maintaining our own repo. Is this PR eligible for merge on a next release soon?

cloudratha:
Strange, I added this quite sometime ago. Did the pull request just get removed?
#1898

kjohnson:
@cloudratha could you pull and re-submit your PR against the `develop` branch? Let's go with the array format, so as to allow for differentiating between possible future metaboxes (and those added by add-ons).

PR #2403, opened 2017-02-03
Fix for Submit Button label not resetting properly

Much2tall:
…ondary (silent) submission from failing to reset the submit button when that submission failed.


Much2tall:
Update: Found out that while the label is getting reset now, it's still not properly allowing for resubmission with Stripe specifically. Any further attempts to resubmit just don't go anywhere.

Much2tall:
Update: Modified the error catching in ninja-forms-stripe to capture the issue as a bandaid fix. An update will be needed in core, but the scope of this issue is completed.

kstover:
Fixed in a recent release.

PR #2420, opened 2017-02-11
Rename date column from 'Date created' to only 'Created'

JeroenSormani:
A small change in the UI

Renaming the column name to only 'Created'. Reason for this the inconsistency in the table head/footer of the list table. The thead/foot were too high due to the long name of the `date` column and the caret being pushed on the second row (because of the max-width of 10% of the 'date' column).

I also tried some other fixes as changing the width, but that caused the timestamp to be on one line, while it being on two lines looks better IMO.

I do think the original 'Date Created' is better in terms of description, but in the end I think 'Created' and a consistent WP UI is better then full name and the inconsistency (but I understand if you feel different, its not really a big deal anyways)

Before:
image

New:
image


kstover:
@JeroenSormani I'm not opposed to making the change, but I'd like to make sure that all of our translations are updated at the same time.

JeroenSormani:
:+1: Anything you'd need from me on this?

kstover:
Do you speak any other languages? :)

JeroenSormani:
Just Dutch and English.
How do you guys normally handle changes in translated texts? I don't suppose keeping it open in a separate branch is the ideal way of handling it..

kstover:
Hey! Sorry for just now getting to this PR!

I'm going to go ahead and merge. We can work on translations from there.

PR #2425, opened 2017-02-14
Fix Checkbox Bug that occurs when using jQuery 1.6+

### LINKED ISSUE #1943: Required checkbox throw "field is required" error even with the checkbox selected when using jQuery v3
Steps to reproduce in v3.0.9:
1. Create a form with a required checkbox
2. Try to submit the form.

h-plus-arry:
Fix #1943: Since Wordpress 4.7 the jQuery Version has changed. The behaviour of the checkbox works not as expected. Because the attr() function does not trigger a state change, anymore.
Older Versions of jQuery took property values sometimes into account when using the attr() function.

As suggested by http://api.jquery.com/prop/ as of jQuery 1.6+ using the prop() function to check the current state of the checkbox instead of attr() fixes that problem.


kstover:
Hey @h-plus-arry. In order to avoid a bunch of conflict overhead, I'm going to close this PR and submit another with just those changes. I really appreciate you putting this PR in; thanks very much!

PR #2435, opened 2017-02-16
Adding a patch to jQuery.val() that triggers a .change() if the targe…

### LINKED ISSUE #2434: jQuery .val() changes for manipulating fields does not bubble properly
Because our backbone listens to .change() events on elements, changes made using jQuery .val() don't bubble properly.

We could patch jQuery.val() in such a way that it fires .change() events if the target is a Ninja Forms field.

kstover:
…t is a Ninja Forms field. Closes #2434.


yavinfour:
Updating a value with pure JS does not update Ninja Forms / Backbone/ Marionette.

```var field = document.querySelector('#nf-field-97');
field.value = 'new value';
```
even when I trigger a change event:

`field.dispatchEvent( new Event('change') );`

PR #2452, opened 2017-02-21
Improve regex patterns

JeroenSormani:
Improve some of the regex patterns to it takes less steps to process.

### DIFF
--- a/includes/Abstracts/MergeTags.php
+++ b/includes/Abstracts/MergeTags.php
@@ -40,7 +40,7 @@ abstract class NF_Abstracts_MergeTags
             return $subject;
         }
 
-        preg_match_all("/{(.*?)}/", $subject, $matches );
+        preg_match_all("/{([^}]*)}/", $subject, $matches );
 
         if( empty( $matches[0] ) ) return $subject;
 
--- a/includes/Actions/Email.php
+++ b/includes/Actions/Email.php
@@ -95,8 +95,8 @@ final class NF_Actions_Email extends NF_Abstracts_Action
             foreach( (array) $email_addresses as $email ){
                 $email = trim( $email );
                 if ( false !== strpos( $email, '<' ) && false !== strpos( $email, '>' ) ) {
-                    preg_match('/(?<=<).*?(?=>)/', $email, $email);
-                    $email = $email[ 0 ];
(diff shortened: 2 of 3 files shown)


kjohnson:
@JeroenSormani can you explain the differences in the RegEx pattern changes, specifically how it benefits over the current implementation? Also, what sort of performance increases are you seeing with this change?

includes/Display/Render.php
```diff
@@ -240,7 +240,7 @@ public static function localize( $form_id )
                 $default_value = apply_filters('ninja_forms_render_default_value', $default_value, $field_type, $settings);
                 if ( $default_value ) {
 
-                    $default_value = preg_replace( '/{.*}/', '', $default_value );
+                    $default_value = preg_replace( '/{[^}]}/', '', $default_value );
```
JeroenSormani: This would be more accurate as it stops at the first encounter of a `}`, while previously it would allow a tag like `{foo}bar}`

includes/Abstracts/MergeTags.php
```diff
@@ -40,7 +40,7 @@ public function replace( $subject )
             return $subject;
         }
 
-        preg_match_all("/{(.*?)}/", $subject, $matches );
+        preg_match_all("/{([^}]*)}/", $subject, $matches );
```
JeroenSormani: This takes less steps to complete

includes/Actions/Email.php
```diff
@@ -95,8 +95,8 @@ protected function check_for_errors( $action_settings )
             foreach( (array) $email_addresses as $email ){
                 $email = trim( $email );
                 if ( false !== strpos( $email, '<' ) && false !== strpos( $email, '>' ) ) {
-                    preg_match('/(?<=<).*?(?=>)/', $email, $email);
-                    $email = $email[ 0 ];
+                    preg_match('/(?:<)[^>]*(?:>)/', $email, $email);
```
JeroenSormani: Also takes less steps to complete

JeroenSormani:
@kjohnson Added some comments, overall it takes less steps to complete / is more accurate. I'm not the best at explaining, but you can see this when you run in a tool like regexbuddy for example

kjohnson:
@JeroenSormani awesome. Thanks for that. I'm going to be reviewing these changes.

PR #2466, opened 2017-02-26
JS Exception Handler

kjohnson:
This adds a global `error` event listener in the `<head />` of front-end pages.

Also, it uses some RegEx to try and identify the soruce plugin, if the exception comes from a script that is enqueue from a plugin directory.

jsexceptionhander_authvsnot


kjohnson:
Currently only shows a notice for administrators, `manage_options`.

PR #2473, opened 2017-02-28
Check for errors after processing/validating fields

### LINKED ISSUE #2417: No way to abort the field process method with an error.
Currently the submission of the form still continues, even if you return errors to the `$data` array, eg.

```php
public function process( $field, $data ) {
    ...
    
    $data['errors']['fields'][ $field['id'] ] = __( 'File upload error' );
    
    return $data;

    ...
}
```

The `maybe_halt()` method is called too late after the actions have been processed, and doesn't look for `_form_data['errors']`. It doesn't make sense to continue the submission and run actions if one of the fields has errored.

You could return errors from the `validate()` field method, but it doesn't make sense to have processing code in there. The `process()` method should have a check to see if errors have been thrown and respond accordingly.

This is specifically required for File Uploads https://github.com/wpninjas/ninja-forms-uploads/issues/203

kjohnson:
Closes #2417.

### DIFF
--- a/includes/AJAX/Controllers/Submission.php
+++ b/includes/AJAX/Controllers/Submission.php
@@ -171,6 +171,11 @@ class NF_AJAX_Controllers_Submission extends NF_Abstracts_Controller
             $this->_data[ 'fields' ][ $field_id ] = $field;
         }
 
+        // Check for errors after validating/processing fields.
+        if( isset( $this->_data[ 'errors' ] ) && $this->_data[ 'errors' ] ){
+            $this->_respond();
+        }
+
         /*
         |--------------------------------------------------------------------------
         | Calculations


polevaultweb:
@kjohnson this doesn't work for fields that set errors to ` $this->_form_data` via https://github.com/wpninjas/ninja-forms/blob/master/includes/AJAX/Controllers/Submission.php#L316. Best to test with FU using this branch, https://github.com/wpninjas/ninja-forms-uploads/pull/204/files and adding `$result = false;` before line 126 https://github.com/wpninjas/ninja-forms-uploads/pull/204/files#diff-0021f3be4cb55dcaf29f2351218ef54cR126

polevaultweb:
@kjohnson ping :)

polevaultweb:
hey @kjohnson any progress on this dude?

Issue #2505, opened 2017-03-06
csv Files being retained after Email Action fires


Much2tall:
Delete csv file from temp directory after Email Action has been processed.

wpnzach:
@Much2tall does this have a PR?  I thought you'd said it was fixed and I just wanted to make sure it got attached to this issue.

PR #2506, opened 2017-03-06
Added a function to delete csv files after Email has been sent.

Much2tall:
Also added logic to remove hr and submit from attached csv files.


Much2tall:
Closes https://github.com/wpninjas/ninja-forms/issues/2505

Issue #2514, opened 2017-03-07
Add a truly, enforced unique number to each form submission


wpnzach:
Sequential numbers have too many issues for this.  We need a way to generate a sequential or random number to ensure uniqueness across all submissions of a form.  Users need this for confirmation purposes (that's the majority use case) so I don't know that sequential is what is necessary, but it might be easiest.

Perhaps instead of tying the submissions ID to the form data, we just check the database for the largest number in that column and increment by one? I don't know.

wpnzach:
I'm marking this urgent as for such a small feature it generates a large amount of support requests because of the confusion it creates for users to not have this enforced as unique.

PR #2524, opened 2017-03-09
Prevent Submission IDs from resetting

Much2tall:
Updated caching to remove _seq_num if it exists from old data. This should prevent the reset of the sequential submission IDs.


Much2tall:
Closes https://github.com/wpninjas/ninja-forms/issues/2514

Issue #2525, opened 2017-03-09
CSV and PDF files written as attachments not properly ordering fields


Much2tall:
Requires multi-part and layout & styles

Issue #2495, opened 2017-03-03
Admin Labels not being used in CSV attached to email action


klhall1987:
The field admin labels are not being used in the submission for the CSV that are attached to email actions. 

Replication Steps: 
1. Create a new form. 
2. Add a text box field and submit. 
3. Change the admin label of the text box. 
4. Add an email action and use the "Attach CSV" Setting in the action. 
5. Preview and submit the form, then download the CSV. You'll see the admin label is not present.

jmcelhaney:
https://wpninjas.kustomerapp.com/app/customers/5891f80316141611002efa1d/event/58b7f35618c9f5110012b2ac

jmcelhaney:
https://wpninjas.kustomerapp.com/app/customers/58ad685baf1a5011002593ec/event/58bef50318c9f51100211c58

klhall1987:
https://wpninjas.kustomerapp.com/app/customers/5891f80316141611002efa1d/event/58b7f35618c9f5110012b2ac

kjohnson:
includes/Actions/Email.php line 207

Issue #2348, opened 2017-01-23
Remove HTML Fields from Email Action CSV


kjohnson:

jimmywags:
Any update on this issues? Thx.

PR #2526, opened 2017-03-09
Properly sort Fields in Email Attachments

Much2tall:
Included catch block to update field order on attachment creation, if the filter exists to sort fields.


Much2tall:
Closes https://github.com/wpninjas/ninja-forms/issues/2525

Much2tall:
Added logic to use admin_label in csv creation (if one is set).

Closes https://github.com/wpninjas/ninja-forms/issues/2495

Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2348

includes/Actions/Email.php
```diff
@@ -197,15 +211,32 @@ private function _format_recipient( $type, $email, $name = '' )
...
+        $today = date( $date_format, current_time( 'timestamp' ) );
+        $csv_array[ 0 ][] = 'Date Submitted';
+        $csv_array[ 1 ][] = $today;
 
         foreach( $fields as $field ){
 
             if( ! isset( $field[ 'label' ] ) ) continue;
-            if( 'hr' == $field['type'] ) continue;
-            if( 'submit' == $field['type'] ) continue;
+            if( 'hr' == $field[ 'type' ] ) continue;
```
kstover: This is a stylistic issue, but I think I'd prefer having an array of ignored filetypes defined above like:

$ignore = array( 'hr', 'submit', 'html', etc.) that is filtered with a WordPress filter. Then check in_array() with the current field type inside your foreach. This would allow third-parties to add field types to the CSV ignore list.

Issue #2529, opened 2017-03-13
IE 11 does not support CSS 'clear: initial;'


amt925:
IE 11 does not support the value `initial` for the `clear` property.

SCSS reference here: https://github.com/wpninjas/ninja-forms/blob/master/assets/scss/front-end/st-defaults.scss#L260

```scss
.five-sixths,
    .four-sixths,
    .one-fourth,
    .one-half,
    .one-sixth,
    .one-third,
    .three-fourths,
    .three-sixths,
    .two-fourths,
    .two-sixths,
    .two-thirds {
        clear: initial;
        float: left;
        margin-left: 2.564102564102564%;
        .inside {
            padding: 20px;
        }
        &::after {
            clear: both;
            content: "";
            display: block;
        }
    }
```

See here: https://developer.mozilla.org/en-US/docs/Web/CSS/clear and here: http://caniuse.com/#feat=css-initial-value for documentation.

This could be set to `inherit` or `none` since the 'initial' value is `none` by default.

```scss
.five-sixths,
    .four-sixths,
    .one-fourth,
    .one-half,
    .one-sixth,
    .one-third,
    .three-fourths,
    .three-sixths,
    .two-fourths,
    .two-sixths,
    .two-thirds {
        clear: inherit;
        float: left;
        margin-left: 2.564102564102564%;
        .inside {
            padding: 20px;
        }
        &::after {
            clear: both;
            content: "";
            display: block;
        }
    }
```

amt925:
Closing in favor of https://github.com/wpninjas/ninja-forms/issues/2316

Issue #2316, opened 2017-01-17
Column classes not working in IE11


wpnzach:
one-half and one-half first are the classes used here:

![image](https://cloud.githubusercontent.com/assets/7498743/22037505/5b16590a-dcc5-11e6-887d-efa61c89531a.png)

Works well in other browsers.

Attached forms are examples.  They use MPF as well, but the issues are present regardless of MPF.

[587d4e55bcf4521100876365-nf_form_01_16_2017_contact.txt](https://github.com/wpninjas/ninja-forms/files/712177/587d4e55bcf4521100876365-nf_form_01_16_2017_contact.txt)
[587d4e55bcf4521100876364-nf_form_01_16_2017_property.txt](https://github.com/wpninjas/ninja-forms/files/712176/587d4e55bcf4521100876364-nf_form_01_16_2017_property.txt)

kjohnson:
Another report from [Nicky Myrthue](https://wpninjas.kustomerapp.com/app/customers/58d99a3e8f7a6910002681e4/event/58e25da733a6f71100bf1b02).

jmcelhaney:
https://wpninjas.kustomerapp.com/app/customers/596fdd727c4aca00139347ce/event/596fdd724d1ce50010f0b653

PR #2530, opened 2017-03-14
Update CSS clear property value to work in IE 11 (issue #2316)

amt925:
This will fix the column classes not working in IE 11.

See here: #2316 and here: #2529


kjohnson:
Related to #2539.

kjohnson:
Closing in favor of #2359.

PR #2535, opened 2017-03-14
Updated default submission error catching to handle a slug.

Much2tall:
Updated error handling logic in submitError.js to accept a slug (if one is provided).

Fixes (potential) Stripe error issue.


Much2tall:
Closes https://github.com/wpninjas/ninja-forms/issues/2534

Issue #2555, opened 2017-03-22
Correct PHP Warnings on Publish


Much2tall:
Related to new Field and Field Meta db queries.

PR #2556, opened 2017-03-22
Remove PHP Warnings on Publish

Much2tall:
Updated some of the chunking logic in the INSERT and UPDATE sections to remove php warnings, which could be contributing to long publish times.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2555

Issue #2404, opened 2017-02-06
Every 10 rows, CSV exports re-add headers


wpnzach:
Every 10 rows, CSV exports re-add headers

`"#","Date Submitted","Name","Email","Message"
"76","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"75","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"74","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"73","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"72","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"71","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"70","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"69","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"68","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"67","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"#","Date Submitted","Name","Email","Message"
"66","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"65","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"64","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"63","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"62","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"61","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
"60","02/06/2017","dghdfhjdghad","dfgdfgh@sdfg.com","hsrtygd68iurftyi"
[...]

PR #2557, opened 2017-03-23
Removed Duplicated Labels from All Submission Exports

Much2tall:
Added a catch to ensure Field Labels are only output once in exports of All Submissions.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2404

PR #2561, opened 2017-03-24
Respond with errors if the field processing returns an error.

polevaultweb:
Resolves https://github.com/wpninjas/ninja-forms/issues/2417

The Field process method would have a return statement, eg.

```
$data['errors']['fields'][ $field['id'] ] = __( 'File upload error' );
return $data;
```


polevaultweb:
@kjohnson this fixes https://github.com/wpninjas/ninja-forms/pull/2473.

PR #2564, opened 2017-03-24
Rename Custom action to WP Hook

### LINKED ISSUE #2562: Rename the "Custom" form action
The "Custom" form action's name is misleading and needs to be renamed to reflect the appropriate use case.

kjohnson:
Closes #2562.

There is a lot of confusion for developers extending Ninja Forms when they see the "Custom" Action in the builder. This "Custom" Action was built to trigger a WordPress action hook, however it is a one way data transfer and does not  modify the form data (ie this is not for filters).

Instead, I suggest that we rename the action to "WP Hook" to better communicate its purpose.

### DIFF
--- a/includes/Actions/Custom.php
+++ b/includes/Actions/Custom.php
@@ -32,7 +32,7 @@ final class NF_Actions_Custom extends NF_Abstracts_Action
     {
         parent::__construct();
 
-        $this->_nicename = __( 'Custom', 'ninja-forms' );
+        $this->_nicename = __( 'WP Hook', 'ninja-forms' );
 
         $settings = Ninja_Forms::config( 'ActionCustomSettings' );
 


Much2tall:
I agree that Custom is a bit too broad.
However, I'm not familiar enough with the context to know if WP Hook is the correct replacement.

alexsuhih:
It seems there is a bug in Custom Actions:
according to these guidelines:
http://developer.ninjaforms.com/codex/custom-form-action/
- I've added the Custom Action and the Hook Tag in the text box 'my_ninja_forms_processing' as descibed.
However, this Custom Action does not work at all: the callback function never calls.
The log is showing this error:
PHP Notice:  Undefined index: tag in \wp-content\plugins\ninja-forms\includes\Abstracts\MergeTags.php on line 48

PR #2565, opened 2017-03-24
Rename the Save action to the Submission action

### LINKED ISSUE #2563: Rename the "Save" form action
With the coming release of Save Progress, the naming of the "Save" action will lead to much confusion. Instead, the name of the "Save" action (in core) needs to be renamed to better reflects the functionality.

kjohnson:
Closes #2563.

Similar to #2564, the naming of the "Save" action can create confusion with the introduction of the Save User Progress ass-on.

Instead, the name should represent its relation to the database and submission.


Much2tall:
Seems Legit.

kstover:
I don't know that "Submission" is descriptive enough. I would think we'd need something like: "Store Submission."

Issue #2571, opened 2017-03-27
Create function to grab our install's TLS version


Much2tall:
- [x] Create the function
- [x] Hook into that function for our Get Help message

double3:
@Much2tall -- i know this thread is closed and the code has been merged, but maybe you'll see this and can help. i bought the PayPal Express add-on for Ninja Forms today. after installing and activating, it gave me this error ... "Ninja Forms was unable to get your TLS version. Please contact your host and ask about updating your environment to support TLS 1.2 and HTTP/1.1." I contacted my host (Network Solutions) and they verified that both of those are already correct on my server. yet NF still says it can't detect the version number. they're not sure what to do and it might be a few days before i hear back on the support ticket i just submitted to NF. any ideas while i wait? i gotta imagine that something on the Network Solutions side is preventing NF from pulling the TLS version. because i was able to install the plugin just fine on a different host (SiteGround). ???

PR #2572, opened 2017-03-27
Added a way to get install's TLS version.

Much2tall:
Added function to get the current install's TLS version and added a call to that function to the System Status page.


Much2tall:
Needed for updates to PayPal and Stripe that catch old TLS versions.

Closes: https://github.com/wpninjas/ninja-forms/issues/2571

includes/Helper.php
```diff
@@ -239,6 +239,45 @@ public static function maybe_unserialize( $original )
...
+     * 
+     * Since 3.0
+     * 
+     * @return float OR false
+     */
+    public static function get_tls()
+    {
+        $php_ver = phpversion();
+        // If we have a php version of 5.6 or higher...
+        if( '5.6.0' > $php_ver ) return false;
```
kjohnson: Shouldn't we be using PHP's [version_compare](http://php.net/manual/en/function.version-compare.php) function?

Much2tall: Good catch.

Much2tall: Updated

Issue #2603, opened 2017-04-07
Add a default value to the Submission Limit message box


Much2tall:
We've had several tickets where users had set a Submission Limit (probably not realizing what it actually was), and failed to include a message.

When they hit the limit, this makes it appear as if their Forms 'break'.

PR #2604, opened 2017-04-07
Added default Submission Limit Message

Much2tall:
Added default message for Forms that have reached their Submission Limit.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2603

includes/Config/FormRestrictionSettings.php
```diff
@@ -74,7 +74,7 @@
                 'label' => __( 'Limit Reached Message', 'ninja-forms' ),
                 'width' => 'full',
                 'group' => 'primary',
-                'value' => '',
+                'value' => 'The Form has reached it\'s submission limit.',
```
kstover: Any time that we output static text, we should run it through WordPress translation functions. __( 'string', 'domain' ) will return the translated string, and _e( 'string', 'domain' ) will echo it. Here, we should use __() like the label does above.

Much2tall: Should be corrected now.

Issue #2624, opened 2017-04-13
Checkboxes incorrectly displayed in Submissions and Exports


Much2tall:
Associated with a MailChimp conflict that's causing the checkbox display function to run multiple times.

PR #2625, opened 2017-04-13
Fix MailChimp Conflict

Much2tall:
Resolved a conflict with MailChimp, where checkboxes in Submissions were trying to render multiple times, causing them to always evaluate as checked.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2624

Issue #2541, opened 2017-03-16
In comma decimated locals, merge tags show wrong calc values in actions


wpnzach:
This is especially prevalent in email actions.

To reproduce, create a product field with a non-zero cent value, i.e. 5.80

In the on form calculations, everything appears properly.  In the email action, however, the calc merge tags will not show the proper value.

PR #2632, opened 2017-04-17
Calculation Refactor

### LINKED ISSUE #2451: Improving Calcs
- [x] Context for Merge Tags - ":calc" on list fields or pulls value (fix context so that it always pulls 'calc' if it exists, then value if calc is unset)
- [x] Rounding setting to calculations (add option of round to x amount of decimal points beneath calc label in builder)
- [x] Save calculations as extra value and show them on submission editing screen and export with submissions
- [x] Remove math.min.js and math.min.js.map and replace its function calls with vanilla js
- [x] Calculations should be available in all merge tags contexts within emails and actions
- [x] Apply padding to calculation content box to avoid overlapping with the merge tag icon
- [x] Calculation content box needs to have key tracking for merge tags it contains

Much2tall:
Closes #2451 .


kstover:
@Much2tall When you edit a calculation by adding/editing the decimal setting, an error is being applied to the settings of the calculation that causes red lines to appear underneath them.

kstover:
@Much2tall If the "decimals" setting is empty, assume 2 decimal places rather than 0.

Much2tall:
@kstover Both issues resolved.

Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2541

PR #2638, opened 2017-04-18
Add option for invisible reCAPTCHA; Close #2551

### LINKED ISSUE #2551: Add support for Google's new Invisible reCAPTCHA
Google just recently released their new Invisible reCAPTCHA, which aims to detect spam bots without any user interaction whatsoever. It looks really cool from a UX standpoint, and I'd love it if you guys could get this integrated in to your plugin. I think a setting when inserting the reCAPTCHA for "Visibility" would probably work great. Something like:

Visibility:

- Visible
- Invisible

The documentation for this is located here: https://developers.google.com/recaptcha/docs/invisible

JacobDB:
This simply adds support for Google's new Invisible reCAPTCHA. I've uploaded a demo [here](https://framework.jacob.rocks/contact-page/).

https://developers.google.com/recaptcha/docs/invisible


JacobDB:
So this isn't actually working I guess, which is why I renamed it to "WIP." It appears that I need to call the function `grecaptcha.execute();` at some point; does anyone have a suggestion on where that should go? Typically it's on the submit button.

JacobDB:
I'm thinking it needs to go in the logic that validates the form, but I'm having a hard time locating that...

JacobDB:
Ah, figured it out. I just placed it in a `try { ... } catch { ... }` just after calling `new nfRecaptcha();`. If this isn't the best place for this, please let me know how I can improve it.

kjohnson:
Hey @JacobDB, thanks for the contribution!

We are currently preparing a major release, but we'll take a look at this on the next pass.

JacobDB:
Sounds good, thanks ☺️

JacobDB:
I just saw the following notice when setting up a new reCAPTCHA key:

> You agree to explicitly inform visitors to your site that you have implemented the Invisible reCAPTCHA on your site and that their use of the Invisible reCAPTCHA is subject to the Google [Privacy Policy](https://www.google.com/intl/en/policies/privacy/) and [Terms of Use](https://www.google.com/intl/en/policies/terms/).

Invisible reCAPTCHA does actually have one visible element, in the lower right corner of the screen. I'm not sure if that's sufficient to meet the above requirement or not; if it's not, I'll need to figure out some way to automatically including a note when a user uses an invisible reCAPTCHA. Perhaps just something in the template.

JacobDB:
One more thing, on the default theme (twentyseventeen) it's showing the field label; I'm thinking that should probably be hidden if visibility is set to hidden. Thoughts?

kstover:
Hey @JacobDB. Sorry for taking so long to get back to this! I think it should have a hidden label. If you want to take care of that and the conflicts, we can get this tested and merged ASAP!

JacobDB:
@kstover, no problem, I'm sure you guys are busy :) I've fixed the conflicts, but I'm having a hard time tracking down how to hide the label. Any suggestions? My thought is to find where the field's render classes are stored and just add `label-hidden` to it if it's visibility is set to invisible.

JacobDB:
@kstover I believe I found the right place to hide the label, please take a look when you have a spare moment and let me know if I need to make any other changes :)

juslintek:
@kjohnson I think you should fix your CI, it cannot find rake file. :-)

kjohnson:
@juslintek we prefer to open new issue for additional features, in trying to keep with a one-to-one system (where possible).

Also, we have a developer actively implementing the CI feature, please excuse the mess.

Issue #2528, opened 2017-03-13
Product fields with in multiples of 1,000 are reduced to a single value


klhall1987:
Product fields that are multiples of 1,000 are reduced to a single digit value. For example 1,000 is shown as 1 in submissions. 10,000 is shown as 10 in submission so on and so forth. 

KR said the issue maybe coming from one of the following lines. 
includes/fields/products.php line 60, 72, and 109

andreasupftw:
This is probably an issue deriving from i18n or rather lack thereof.
It's 1 AM here, so I can't really dive into it. Just wanted to chime in, or something...

andrewfleming:
This could also be a result of [PHPs string conversion to numbers](http://php.net/manual/en/language.types.string.php#language.types.string.conversion), and the quantity field formatting numbers with a comma thousand separator. 

I'm assuming the quantity is saved in the DB as a string, so when it is used in a mathematical equation PHP converts the string to a number following the rules in the link above and everything after the thousand separator is disregarded. 

[Lines 60 -66 of /includes/fields/products.php](https://github.com/wpninjas/ninja-forms/blob/97377f16fe0a552cb0964a0ce09bc1837d35300c/includes/Fields/Product.php#L60-L66) are an examples of this. `floatval` disregards whatever is after the comma thousand seperator too.

The solution I went with when I ran into a similar issue to this was to filter values before they were saved to the DB with `filter_var( $number, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION );`

PR #2656, opened 2017-04-25
Fixed product total output

Much2tall:
Updated logic in Products to bandaid-fix merge tag output of product totals.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2528

includes/Fields/Product.php
```diff
@@ -106,7 +107,10 @@ public function filter_required_setting( $field )
 
     public function merge_tag_value( $value, $field )
     {
-        $product_price = preg_replace ('/[^\d,\.]/', '', $field[ 'product_price' ] );
+        // TODO: Replaced this to fix English locales.
+        // Other locales are still broken and will need to be addressed in refactor.
+//        $product_price = preg_replace ('/[^\d,\.]/', '', $field[ 'product_price' ] );
+        $product_price =  str_replace( array( ',', '$' ), '', $field[ 'product_price' ] );
```
kstover: We should know the actual thousands separator and currency symbol being used. Can't we just grab those here instead of assuming English?

includes/Fields/Product.php
```diff
@@ -57,7 +57,8 @@ public function process( $product, $data )
             $related[ $type ] = &$data[ 'fields' ][ $key ]; // Assign by reference
         }
 
-        $total = floatval( $product[ 'product_price' ] );
+        $total = str_replace( array( ',', '$' ), '', $product[ 'product_price' ] );
```
kstover: We should know the actual thousands separator and currency symbol being used. Can't we just grab those here instead of assuming English?

Much2tall: @kstover We can grab those and replace them. The question is whether or not that change is within the scope of this fix. Ideally, we would never save the price as a pre-formatted string (as we already are), since this makes doing math on it in any way a nightmare (which is what led to this fix in the first place).

It would be more efficient, in my opinion, to bugfix this now, and during the product refactor (whenever that comes up) go back over it and write a method that formats prices for display only. That way, we always save them as a standard locale format and can perform math on them before rendering.

Much2tall:
Bugfix does not introduce any additional issues for non-English locales. It just doesn't correct them.

PR #2693, opened 2017-05-02
Add support for hypens in merge tags/field keys.

### LINKED ISSUE #2691: New Merge tag system is stripping field keys with hyphens
If a field key contains a hyphen, then everything before the hyphen is removed from the field key.

kjohnson:
Closes #2691.


klhall1987:
Fix tested and working correctly. Merging into develop.

Issue #2688, opened 2017-05-02
Refactor forms endpoint to query on demand


kjohnson:
On the Forms Controller (`includes/Database/FormsController.php`), the DB is being queried in the constructor.

Instead, it should be queried on demand, so as not be hitting the DB on every page load.

PR #2704, opened 2017-05-02
Updated location of query to avoid loading on construct.

Much2tall:


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2688

Issue #2711, opened 2017-05-04
Checkbox Lists improperly displaying in Submissions


Much2tall:
For conditionally hidden Fields

PR #2712, opened 2017-05-04
Broken Checkbox Lists in Submissions

Much2tall:
Added array check to sweep out invalid values from conditionally hidden Checkbox List Fields.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2711

Much2tall:
Should work properly now.

kstover:
Well, while looking into this bug, I found another with editing submissions with checkbox lists within the admin.

If you edit the submission by unchecking every option, your changes aren't saved. You can uncheck everything but one, and it will save. When you try to save with NO options checked, it will return to its original state.

Much2tall:
@kstover 
I did some checking, and that issue also appears on the master branch.

Much2tall:
@kstover 
Issue should now be resolved.

Issue #2726, opened 2017-05-09
Sort forms by title by default in the admin list of forms


isleshocky77:
Currently, I believe, it sorts the forms by id which since they're no longer even shown is confusing. It would be much more helpful to have them sorted by title by default.

PR #2736, opened 2017-05-12
Sort list of forms by title by default.

isleshocky77:
This is the sane thing to do since by default most places show the title of the form first and is most prevalent and used.

for #2726


kjohnson:
Thanks, @isleshocky77 .

This is something that we will need to discuss internally, as it is a change in convention in the product.

isleshocky77:
@kjohnson Completely understand. It's just very frustrating to have it sorted by id. Previously, it sort of made sense the first column always was the form id; however, you've removed that column.  It just makes more sense to have the forms sorted by title first everywhere.

I would add that I would like to see the drop-down on the submissions page also sorted by title. Would you have an issue with that?

PR #2753, opened 2017-05-22
Updating to remove php short tag.

daviddrace:
Noticed this while debugging a Wordpress site running on a server with php short tags disabled. 

Short tags should be avoided, since using them requires a specific PHP configuration to work. See http://php.net/manual/en/language.basic-syntax.phptags.php.


kjohnson:
Thanks for the catch, @daviddrace !

Issue #2684, opened 2017-05-02
Submit button does not re-enable once validation issues are fixed


alijaza:
I have reproduced this in two ways:

* create a simple for with just a single antispam field and a submit button. Enter the wrong input and submit. Then correct the input and try to submit again. Submit button is disabled.

* Second way is similar but slightly different. I have a field with custom field validation logic. If I enter the wrong input and an error is shown, I cannot submit again even if I correct the input.

Version 3.0.34.1, WP 4.7.4

jmcelhaney:
@kjohnson I was able to replicate the issue with the Anti-spam field.

tan-ahmed:
I too was able to replicate this issue!

PR #2759, opened 2017-05-24
Allow resubmission after changing value of anti-spam

Much2tall:
Updated error reporting for anti-spam field so that it can clear the error on keyup.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2684

PR #2786, opened 2017-06-09
Keep context of form model on field collection reset

Much2tall:
Corrected an issue in default filters that was removing form model as an option of the field collection. Added a radio message to the field collection on reset.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2785

PR #2787, opened 2017-06-09
Minified javascript

wickedpuppy:
I minified the javascript to prevent autop from adding extra <p></p> lines inside the script and resulting in errors and plugin broken when used in many templates and plugins filtering the_content.

Consider moving it to template file ASAP


kstover:
I think that this is as a good solution for the time being; a front-end refactor is on our roadmap, which would take are of this issue.

PR #2792, opened 2017-06-13
Init NF_Session as needed

### LINKED ISSUE #2579: Sessions are still being regenerated on every page
This is a long-standing issue. See #954, #1502, #1236, #1649.

I previously submitted a PR which fixes this: https://github.com/wpninjas/ninja-forms/issues/1803 

The PR and all issues are closed, but this is still not fixed. 

You can reproduce the problem quite easily: 
```
add_action( 'init', function() {
    if ( class_exists('WP_Session') ) {
        echo WP_Session::get_instance()->session_id;
    }    
});
```
Reload the page a few times and you'll see the session ID is different on every page load. That effectively makes the session entirely pointless. 

Worse, you're using the `WP_Session` library, which is used by other plugins too. The way `WP_Session` works is that every installed plugin that uses it will share the same session. From your own code: 

```
if ( ! defined( 'WP_SESSION_COOKIE' ) ) {
   define( 'WP_SESSION_COOKIE', 'nf_wp_session' );
}
```

https://github.com/wpninjas/ninja-forms/blob/master/includes/Session.php#L46-L48

You'll see that this is the exact same way that other plugins who use `WP_Session` are defining the constant: 

- [Easy Digital Downloads](https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/class-edd-session.php#L84-L86)
- [Charitable](https://github.com/Charitable/Charitable/blob/master/includes/public/class-charitable-session.php#L49-L51)
- [Give](https://github.com/WordImpress/Give/blob/master/includes/class-give-session.php#L96-L99)

In other words, Ninja Forms, Charitable, Easy Digital Downloads, Give and just about anyone else who uses `WP_Session` in their plugin is going to share the same session. 

The reason your implementation of `WP_Session` is breaking things is because you are _not_ setting the cookie on page load: 

screenshot_2016-09-15_12_53_58

As a result, until some action happens to take place that involves Ninja Forms setting something on the session, Ni

kjohnson:
Closes #2579.

`NF_Session` is only used by the `resume` feature of form submission.

Instead of initializing `NF_Session` on `plugins_loaded`, initialize as needed.

### DIFF
--- a/includes/Session.php
+++ b/includes/Session.php
@@ -57,7 +57,8 @@ class NF_Session {
         add_filter( 'wp_session_expiration_variant', array( $this, 'set_expiration_variant_time' ), 99999 );
         add_filter( 'wp_session_expiration', array( $this, 'set_expiration_time' ), 99999 );
 
-        add_action( 'plugins_loaded', array( $this, 'init' ), -1 );
+        // Since we only loading this as needed, we will need to call init() manually.
+        // add_action( 'plugins_loaded', array( $this, 'init' ), -1 );
 
     }
     /**
--- a/ninja-forms.php
+++ b/ninja-forms.php
@@ -291,8 +291,6 @@ if( get_option( 'ninja_forms_load_deprecated', FALSE ) && ! ( isset( $_POST[ 'nf
                  */
                 self::$instance->_eos[ 'parser' ] = require_once 'includes/Libraries/EOS/Parser.php';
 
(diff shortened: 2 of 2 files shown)


kstover:
@jmcelhaney Can you test this PR with PayPal?

kstover:
@jmcelhaney Confirmed that this works with PayPal Express.

Issue #2733, opened 2017-05-12
Calculations decimal setting is not being respected in actions


klhall1987:
If the user has the decimal value set to 0 in a calculation, it is not being respected in some actions. I tested in emails and in success messages. 

Screenshots:
Calc Setting: 
https://goo.gl/2oTjgV

Success message: 
https://goo.gl/kWmImm

klhall1987:
This is still an issue.

PR #2812, opened 2017-06-19
Prevent 0 decimal value from defaulting to 2 decimal places after submit.

Much2tall:
Modified decimal value check to evaluate >= 0 instead of not false.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2733

PR #2813, opened 2017-06-20
Use Year Range setting on Date Field to limit the dates selected

### LINKED ISSUE #2484: Date Field does not respect Year Range advanced setting
Reproduction steps:

1) Add a Date field to a form
2) In the Year Range option under Advanced Settings of that Date field, set Start Year to 2017 and End Year to 2018.
3) On the front end with this form, you can still select dates that are outside of the indicated Year Range.

dbeja:
I used 1st January of Start Year as mindate and 31st December of End Year as maxdate so that Year Range advanced settings on Date Field limit the dates that can be selected.

Fixes #2484.


kstover:
In my testing, I set the range from 2017 to 2017. This left only a few dates selectable, although it DOES limit the years I can view properly.

Am I setting this up wrong?

dbeja:
I tried now setting that range (2017-2017) and I could select any date from 1st January 2017 until 31st December 2017. Did you made any other selection?

By the way, two questions about this pull request:

1. I only committed the changed javascript file, I didn't committed the minified javascript file generated by gulp. Is that correct? Or should I commit also those files?
2. This file is using spaces instead of tabs. Should I take the chance to convert it to tabs?

Thanks

dbeja:
I did more tests with my changes and couldn't replicate your issues.
You just set the range from 2017 to 2017 and you can't select all the dates from January 1st 2017 until December 31st 2017?

Thanks!

kstover:
Just tested again, and it appears to work for me.
I'm going to merge.

Issue #2725, opened 2017-05-09
Calcs need to scrub line breaks from their equations


Much2tall:

Much2tall:
Reopening because a user was able to reproduce this bug on newest version.

PR #2819, opened 2017-06-22
Remove Whitespace from Calcs

Much2tall:
Made another pass at correcting whitespace issues. Added try/catch blocks to every instance of the math library being called.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2725

PR #2825, opened 2017-06-26
Fix addon modals.

Much2tall:
Cast FormID back to a string before localization to resolve issues with addon modals.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms-stripe/issues/86

PR #2830, opened 2017-06-30
Show total quantity on Submissions list when using products with inline quantites

### LINKED ISSUE #2527: Product field in line quantity fields do not appear in submissions
Product field inline quantity fields do not appear in submissions. This makes site admins have to calculate how many products a user purchases.

dbeja:
Fixes #2527.

Product inline quantities are saved inside product field value and the submissions columns are generated from the form fields. As a form doesn't have the quantity field (when it's using inline quantities) what I did was:

1. When creating the NF Submissions columns headers I check if there are any product type fields and those products have inline quantity setting activated. If that is true, I manually add a column header after the last product with ID **nf_inline_quantity**
2. When adding content to each row column of Submissions list, if the column is **nf_inline_quantity** I get all form fields that are product type with inline quantity and sum all their values from the current submission

### DIFF
--- a/includes/Admin/CPT/Submission.php
+++ b/includes/Admin/CPT/Submission.php
@@ -144,6 +144,9 @@ class NF_Admin_CPT_Submission
 
         $form_fields = Ninja_Forms()->form( $form_id )->get_fields();
 
+        $has_inline_quantities = false;
+        $last_product_id = 0;
+
         foreach( $form_fields as $field ) {
 
             if( is_object( $field ) ) {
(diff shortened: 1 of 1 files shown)


kstover:
@dbeja Thanks for the PR! I don't know that this works as a solution because:

- It adds all of the products inline quantities together. If I have multiple products, then I see all the quantities added together on the submission list. This doesn't seem very useful.
- It doesn't affect the export or submission edit screen.

I think that a solution to this problem would need to show the "total" of the item in the submission list and exports, while showing quantity and product price on the edit screen.

kstover:
Closing in favour of: https://github.com/wpninjas/ninja-forms/pull/2881

Issue #2219, opened 2016-12-19
'disable_input' not working for email type field


gvanto:
Hooking into filter as follows, 'disable_input' works for 'name' (text) field but not for email field:

```
add_filter('ninja_forms_display_fields', 'dpcontent_nf_display_fields', 10, 1);

function dpcontent_nf_display_fields($fields) {

if (is_user_logged_in()) {
            // If user logged in, disable inputs for Name + email
            // https://wordpress.org/support/topic/how-to-omit-fields-programmatically-in-nf3/ 
            for ($i = 0; $i < count($fields); $i++) {
                if (array_key_exists('admin_label', $fields[$i])) {
                    if ($fields[$i]['admin_label'] == 'email' /* 'name' */) {                        
                        $fields[$i]['disable_input'] = 1;
                        // this works on name but not email field                        
                    }
                }                
            }            
        }
```

kstover:
@gvanto Are you working on the 3.0 UX?

gvanto:
@kstover yes NF 3

kstover:
This should be taken care of soon, as it's an easy fix.

PR #2849, opened 2017-07-06
Email Field not honoring Disable Input

Much2tall:
Updated the template for Email Fields to allow for programatic disabling of input.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms/issues/2219

PR #2852, opened 2017-07-07
Add mask for credit card expiration.

Much2tall:
Added default mask to creditcardexpiration field to prevent invalid input for Stripe.


Much2tall:
Closes: https://github.com/wpninjas/ninja-forms-stripe/issues/84

Issue #2648, opened 2017-04-21
reCaptcha - Submitting form without validating locks the user from submitting.


MongolianMisfit:
As the title suggests, submitting without validating reCaptcha locks the form with an error, rendering the user unable to submit the form. 

The only solution in this case is a complete reload of the page, and resubmission whilst appropriately validating reCaptcha.

patricedefago:
Hello,

Same bug for me.

It's not good for conversion.

Thanks to correct this.

Have a good day☀️🍀😀

kstover:
I cannot reproduce this on the current version of Ninja Forms. If this is still an issue, please feel free to re-open this issue.

klhall1987:
@kstover Here are the replication steps: 

1. Create a new form with a checkbox, recaptcha field and submit button.
2. Preview the form. 
3. Click the submit button. 
4. Validate the Recaptcha Field and try to resubmit the form.

roborourke:
@kstover same here, if you fill out all fields except recaptcha and submit it locks the form for resubmission.

Additionally the error message doesn't show, it appears that `id` isn't passed when generating the field's error div.

PR #2869, opened 2017-07-14
Submit disabled on failed reCaptcha

Much2tall:
Modified error logic for reCaptcha so that it now properly handles a failed submission.


Much2tall:
Closes https://github.com/wpninjas/ninja-forms/issues/2648

PR #2880, opened 2017-07-18
Actions now listen for field deletion and remove any attributes that …

### LINKED ISSUE #2840: Deleted fields aren't being tracked in action settings
Fields used in action settings are showing up as the field merge tag after they are deleted. For example if in User Management you set a Password field as the Password Field setting, then delete the password field, then the password will be set to {field:password} when the action is processed. 

Replication Steps(Using the user management example): 
1. Create a form using the Update Profile template.
2. Delete the password field. 
3. add a dump/die statement to include the Key and Value on like 200 in user_management/includes/actions/UpdateProfile.php. 
4. Fill out the form and view the AJAX request that is sent.

kstover:
Actions now listen for field deletion and remove any attributes that reference that field merge tag. Closes #2840.


assets/js/builder/controllers/actions/deleteFieldListener.js
```diff
@@ -0,0 +1,40 @@
...
+			actionModel.listenTo( nfRadio.channel( 'fields' ), 'delete:field', this.maybeUpdateSettings );
+		},
+
+		maybeUpdateSettings: function( fieldModel ) {
+			_.each( this.attributes, function( attr, key ) {
+				var fieldMergeTag = '{field:' + fieldModel.get( 'key' ) + '}';
+				if ( _.isString( attr ) ) {
+					this.set( key, attr.replace( fieldMergeTag, '' ) );
+				} else if ( _.isArray( attr ) ) {
+					var index = attr.indexOf( fieldMergeTag );
```
kjohnson: If an `attr` is an array, then the check needs to be nested, checking for a possible string position inside of each item in the array.

kstover:
Should now work with option repeaters.

assets/js/builder/controllers/actions/deleteFieldListener.js
```diff
@@ -0,0 +1,47 @@
...
+		},
+
+		maybeUpdateSettings: function( fieldModel ) {
+
+			_.each( this.attributes, function( attr, key ) {
+				var fieldMergeTag = '{field:' + fieldModel.get( 'key' ) + '}';
+				if ( _.isString( attr ) ) {
+					this.set( key, attr.replace( fieldMergeTag, '' ) );
+				} else if ( _.isArray( attr ) ) {
+					_.each( attr, function( obj, index ) {
```
kjohnson: Not seeing this work for the Option Repeater.

webhooks

calculations

kstover: Should be working now.

PR #2881, opened 2017-07-18
Products (with Inline Quantities) in the Submissions Table

### LINKED ISSUE #2527: Product field in line quantity fields do not appear in submissions

kstover:
Modifying the way that inline product values are shown on the submissions table and editor. The sub table now reads: $10.40 ( 3 ) and the editor shows price, quantity, and total. Closes #2527.


kjohnson:
@kstover I have a concern about simply pulling the Product Price directly from the (current) published state of the form. If the form is modified after the time of submission, the recorded submission changes, which creates a discrepancy with the recorded "Total".

I would suggest that we store the Product Price (at the time of the submission) as "Extra" submission data, which can be used to display the submission. This would, in turn, raise a question about backwards compatibility with previously recorded submissions, for which I would suggest not showing this format (as opposed to using the live form, which can cause the same issue and is essentially a guess).

**Original Form (at time of submission)**

before

**After the Form was updated to reflect a pricing change.**

after

kstover:
Now stores a copy of the historical price for display when editing a submission with products.

kstover:
Updated to reflect historical prices in submissions table.

PR #2896, opened 2017-07-20
1st pass front end a11y

roborourke:
I've done some preliminary testing with VoiceOver on mac - toggle it using command+f5.

It's quite chatty but seems ok. Ideally it should be tested by someone who's far more used to it but this is certainly a step in the right direction.

Also you should double check I haven't broken anything adding in the `aria-invalid` attribute behaviour!

Props to @rianrietveld for the advice.


roborourke:
Oh I was thinking about adding the `aria-required` attribute too but since it's announcing required fields via the inline error message it's not super necessary.

And before I forget I couldn't find where the file upload progress count is updated - ideally the `aria-valuenow` attribute will be updated with the percentage too.

roborourke:
Closed in favour of #2899

PR #2898, opened 2017-07-21
Fixes created_at on duplicated forms

### LINKED ISSUE #2060: Duplicating a form should set the created date to today's date

Dwarve:
Updates the created_at field in the database for each element of the
form that is duplicated to reflect the current date/time. Also slight JS
update to handle the immediate display as well. Closes #2060


kstover:
Hmm. I'm not seeing the JS change. When I duplicate, I don't see an updated created date unless I reload the dashboard. @Dwarve

PR #2914, opened 2017-07-25
Collect payment refactor

### LINKED ISSUE #2913: Refactor collect payment action
Refactor the collect payment action to be more user friendly.

klhall1987:
We have created a new convention for the collect payment action. The action is set up to allow users to select the type of payment they would like to use and a value. 

closes #2913,


Much2tall:
Breaks some old Collect Payment Actions that are using Fields not defined in the new list.
ie (checkboxlist_23144603532:calc) will be removed upon update, and the same value cannot be selected from the Field list, as Checkbox List Fields aren't included in the new settings.

kstover:
@Much2tall Unknown fields are now added to the list, so Radio lists, etc. will show up as one-offs if the field was already setup as the total.

PR #2922, opened 2017-07-27
Sets created_at on all Imports

### LINKED ISSUE #2837: Templates and Imported Forms Should Have Current Date as Created Date
When you import a form or click on a template from the "Add New" widget, the creation date is hard-coded in the form. We should update that to the current date/time of the user's WordPress install.

Dwarve:
Import function updated to set all created_at fields on imported forms,
fields, and actions. Closes #2837.


Much2tall:
@Dwarve 
The created_at should only be updated on new imports.
Anything that falls under is_conversion should keep its original created_at setting.

Dwarve:
@Much2tall Added the conversion check.

PR #2958, opened 2017-08-03
Locale-based language selection for reCAPTCHA

mazimmer:
We use ninja-forms in our multi-lang website and needed a way for the reCAPTCHA to use the language defined in the locale, as set by the current anonymous user.
So we use the keyword "locale" in the reCAPTCHA settings and, if defined, set the language accordingly when rendering.


kjohnson:
@mazimmer does reCAPTCHA support every language in the `locale`? This might be a good place to add a filter, which could then be leveraged with custom code.

kjohnson:
@marcoazimmer after some additional consideration, I can accept a filter, but would rather not bundle this functionality into the core plugin.

PR #2978, opened 2017-08-08
Update gutters after domain change.

### LINKED ISSUE #2977: Gutter update timing issue with domain change.
Related to wpninjas/ninja-forms-multi-part#104

Steps to Reproduce:
- Add fields to form.
- Change Domain to "Emails & Actions".
- Open an Action's settings drawer.
- Change Domain to "Fields" (without closing drawer).

The timing of the domain change, and the resulting drawer close, causes a visual artifact in which the gutter remains atop the field list.

kjohnson:
Closes #2977.


kstover:
@klhall1987 @Much2tall This will need to be tested with Layout & Styles and Multi-Part Forms, as both of those make use of the gutter system.

PR #3024, opened 2017-09-02
Add throttled check to THREE upgrade notice.

kjohnson:


kstover:
Assigning this back to @kjohnson so that he can implement a 1-100 percentage, along with an initial 5% implementation.

PR #3044, opened 2017-09-07
License Activation Error Notice

### LINKED ISSUE #3042: Update EDD Software Licensing Implementation
Our licensing API returns 'failed' rather than the API error itself. This needs to be updated.

kstover:
Adding a notice to the settings screen when a license activation error occurs and the nf_debug querystring is set to 1. Closes #3042.

I had to update the Notices.php file because it was returning early when there was an issue while looping over admin notices instead of continuing. This made it impossible to fire later notices when an early one should have been skipped.

I also added a flag for admin notices called "ignore_spam" which tells the admin notice class you want the notice to display, regardless of how many Ninja Forms notices there are on the screen.

### DIFF
--- a/includes/Admin/Notices.php
+++ b/includes/Admin/Notices.php
@@ -60,11 +60,17 @@ class NF_Admin_Notices
         }
 
         foreach ( $admin_notices as $slug => $admin_notice ) {
-            // Call for spam protection
-            if ( $this->anti_notice_spam() ) {
-                return false;
+
+            if ( isset ( $admin_notice[ 'ignore_spam' ] ) && true == $admin_notice[ 'ignore_spam' ] ) {
+                $ignore_spam = true;
+            } else {
+                $ignore_spam = false;
             }
 
+            // Call for spam protection
+            if ( ! $ignore_spam && $this->anti_notice_spam() ) {
+                continue;
+            }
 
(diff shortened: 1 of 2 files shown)


includes/Admin/Notices.php
```diff
@@ -138,6 +146,8 @@ public function admin_notice( $admin_notices ) {
                 }
             }
         }
+
+        // die( 'done looping' );
```
kjohnson: Stray comment.

includes/Admin/Notices.php
```diff
@@ -74,10 +80,12 @@ public function admin_notice( $admin_notices ) {
                 if( ( isset( $admin_notices[ $slug ][ 'blacklist' ] ) && $this->admin_notice_pages_blacklist( $admin_notices[ $slug ][ 'blacklist' ] ) )
                     || ( isset( $admin_notices[ $slug ][ 'pages' ] ) && ! $this->admin_notice_pages( $admin_notices[ $slug ][ 'pages' ] ) )
                 ) {
-                    return false;
+                    continue;
                 }
             }
 
+            
+
```
kjohnson: Extra spacing?

kjohnson:
Closes #3057 .

PR #3048, opened 2017-09-08
Add a beforeunload notice when converting forms.

### LINKED ISSUE #3045: Stop users from leaving conversion page before conversion process is complete
add a prompt or other warning to the conversion page so the user does not navigate away before conversion is completed.

kjohnson:
Closes #3045 .

### DIFF
--- a/deprecated/upgrade/upgrade.js
+++ b/deprecated/upgrade/upgrade.js
@@ -95,6 +95,7 @@ jQuery(document).ready(function($) {
         },
 
         start: function () {
+
             _.each( nfThreeUpgrade.forms, function( formID ) {
                 this.forms.push({
                     id: formID,
@@ -119,6 +120,11 @@ jQuery(document).ready(function($) {
             console.log( 'HERE' );
             console.log( app );
             app.step = 'converting';
+
+            // Add a notice if the user tries to navigate away during conversion.
+            jQuery( window ).bind( 'beforeunload', function(){
+                return 'You have unsaved changes.';
+            } );
 
             $.post( ajaxurl, { nf2to3: 1, action: 'ninja_forms_ajax_migrate_database' }, function( response ) {
 


deprecated/upgrade/upgrade.js
```diff
@@ -120,6 +121,11 @@ jQuery(document).ready(function($) {
             console.log( app );
             app.step = 'converting';
 
+            // Add a notice if the user tries to navigate away during conversion.
+            jQuery( window ).bind( 'beforeunload', function(){
+                return 'You have unsaved changes.';
```
kstover: This should be a localised string in PHP; I'd also suggest getting @wpnzach's input into what the text of the notice should be.

kjohnson: @kstover I don't think that this text can actually be defined. Rather, the browser determines the actual text.

kstover: Then why return a string?

kjohnson: To my understanding, a message needs to be returned, but is not respected.
Each browsers is different, and can be different across versions.
https://stackoverflow.com/a/38880926

kjohnson: Also, it seems that browsers are removing support for custom text.
https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Browser_compatibility

kjohnson: > When this event returns (or sets the returnValue property to) a value other than null or undefined, the user is prompted to confirm the page unload. In some browsers, the return value of the event is displayed in this dialog. Starting with Firefox 4, Chrome 51, Opera 38 and Safari 9.1, a generic string not under the control of the webpage will be shown instead of the returned string.

kstover: I see that this is the same thing we're doing already on the builder. I think that's fine. Merging away!

PR #3058, opened 2017-09-11
Fixed a bug with blacklisting blocking future notices.

### LINKED ISSUE #3057: Re-show optin banner in THREE
The `opt-in` notice is not showing.

kjohnson:
The opt-in notice was not being shown because the blacklist check of a previous notice was early returning, instead of continuing the loop.

Closes #3057 .


kstover:
This is addressed in this PR: https://github.com/wpninjas/ninja-forms/pull/3044/

If we merge this PR, it'll create merge conflicts when looking at the one I mentioned above.

@kjohnson Thoughts?

kjohnson:
This seems moot, with #3044 .

PR #3060, opened 2017-09-11
Group and Prioritize Add-ons list.

kjohnson:
- Adds category grouping to the add-ons list (feed).
- De-prioritizes the order of already installed add-ons.
- De-prioritizes the CRM or Newsletter groups if a grouped add-on is already installed.

### DIFF
--- a/includes/addons.json
+++ b/includes/addons.json
@@ -1,0 +1,365 @@
+{
+  "addons": [
+    {
+      "title": "Layout and Styles",
+      "image": "assets/img/add-ons/layout-styles.png",
+      "content": "Easily create multi-column form layouts and beautifully styled form elements without mastering CSS.",
+      "link": "https://ninjaforms.com/extensions/layout-styles/?utm_medium=plugin&utm_source=plugin-addons-page&utm_campaign=Ninja+Forms+Addons+Page&utm_content=Layout+and+Styles",
+      "plugin": "ninja-forms-layout-styles/ninja-forms-style.php",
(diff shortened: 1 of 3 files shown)


kjohnson:
| Core (Only) | File Uploads (Builder) | MailChimp (Newsletter) |
| --- | --- | --- |
| image | image | image |

includes/addons.json
```diff
@@ -0,0 +1,373 @@
+{
+  "addons": [
+    {
+      "title": "Layout and Styles",
+      "image": "assets/img/add-ons/layout-styles.png",
+      "content": "Easily create multi-column form layouts and beautifully styled form elements without mastering CSS.",
+      "link": "https://ninjaforms.com/extensions/layout-styles/?utm_medium=plugin&utm_source=plugin-addons-page&utm_campaign=Ninja+Forms+Addons+Page&utm_content=Layout+and+Styles",
+      "plugin": "ninja-forms-layout-styles/ninja-forms-style.php",
+      "docs": "https://ninjaforms.com/docs/layouts-and-styles/?utm_medium=plugin&utm_source=plugin-addons-page&utm_campaign=Ninja+Forms+Addons+Page&utm_content=Layout+and+Styles+Docs",
+      "version": "3.0.18",
```
Much2tall: Who do these include version numbers?
Will need to be maintained in future updates if we leave them. (Duplicated work.)

kjohnson: The feed was originally going to be sourced from the NinjaForms server, so the idea was that we could ID updates - even without license keys. However, the feed was added to the plugin, so this needs to be revisited.

Much2tall:
Closing in favor of additional changes before resubmission of PR.

PR #3067, opened 2017-09-13
Adds a dispatcher for sending requests to api.ninjaforms.com if the user is opted in.

kstover:
Also adds basic telemetry for:

- gathering server environment information every month.
- sending data when the user opts in and upgrades.

Also adds a modal prompting the user to opt-in if the user is attempting to upgrade and isn't opted in.


deprecated/upgrade/class-submenu.php
```diff
@@ -155,6 +166,150 @@ public function settings_upgrade_button_display()
...
+
+    public function optin() {
+        if ( ! current_user_can( 'manage_options' ) ) return false;
+
+        $api_url = 'http://api.ninjaforms.com/';
+
+        /**
+         * Update our tracking option.
+         */
+        update_option( 'ninja_forms_allow_tracking', true );
```
kjohnson: We also need to clear the `ninja_forms_do_not_allow_tracking` value.
`update_option( 'ninja_forms_do_not_allow_tracking', true );`

kjohnson: We also need to clear the ninja_forms_do_not_allow_tracking value.
update_option( 'ninja_forms_do_not_allow_tracking', true );

deprecated/upgrade/upgrade.js
```diff
@@ -1,5 +1,59 @@
 jQuery(document).ready(function($) {
 
+    /**
+     * Show an opt-in modal if the user isn't currently opted in.
+     */
+    if ( 0 == nfThreeUpgrade.optedIn ) {
```
kjohnson: This should also check if an add-on is installed, and automatically opt-in the user because they are using a premium product.

PR #3075, opened 2017-09-18
Move forced required validation for submission.

### LINKED ISSUE #3069: Required Fields that have not been focused do not get caught as required
Any required field that has been conditionally shown (part or individual field) is not being treated as "required" for Form Submission to take place.

kjohnson:
Required Validation is forced on fields when submitting, via the Form Content Data.
However, with Layout & Styles the Form Content Data is overridden and the forcing `clean` to false - for the purpose of required validation - is not running.

This pull request moves the code from the Form Content Data `validateFields()` method and runs it directly inside of `submit.js`.

Closes #3069 .


assets/js/front-end/controllers/validateRequired.js
```diff
@@ -49,6 +49,7 @@ define([], function() {
 		},
 
 		validateModelData: function( model ) {
+			console.log( model );
```
Much2tall: Should be removed unless there is a specific reason to leave it.

kjohnson: Ah... you got me!

Issue #2974, opened 2017-08-08
Email "to" fields including TO, BCC, CC with multiple merge tags need logic to know when to truncate trailing commas


wpnzach:
It's a not-uncommon use case (two this week) where users might want multiple email addresses in the TO, CC, or BCC fields in the email actions.  The issue is that if merge tags are used for these, but the form is submit without one of these email addresses, the form will fail to submit because of the improper formatting of those fields.

An example use case might be a form registering several people at once for an event.  They should be able to have an email action with multiple recipients, maybe up to four....but if only two are registered instead the merge tags in the CC of an email action might be blank, causing the form submission to fail.

kjohnson:
PR merged. Closing issue.

PR #3088, opened 2017-09-19
Remove extra commas from email address settings.

klhall1987:
Fixes # 2974

- Created a helper method in the email action that removes extra comma separators from email settings(ie To, BCC, CC, reply-to, and from address'). 

Testing Steps:
1. Create a form with multiple email address fields. 
2. Set the email address fields as settings in the TO, BCC, CC, reply-to, or from address as comma 
separated merge tags. 
3. Submit the form and see if the email goes through. 

@wpninjas/developers

### DIFF
--- a/includes/Actions/Email.php
+++ b/includes/Actions/Email.php
@@ -47,6 +47,8 @@ final class NF_Actions_Email extends NF_Abstracts_Action
 
     public function process( $action_settings, $form_id, $data )
     {
+        $action_settings = $this->sanitize_address_fields( $action_settings );
+
         $errors = $this->check_for_errors( $action_settings );
 
         $headers = $this->_get_headers( $action_settings );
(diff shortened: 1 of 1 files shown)


includes/Actions/Email.php
```diff
@@ -103,6 +106,53 @@ public function process( $action_settings, $form_id, $data )
...
+            if( is_array( $action_settings[ $setting ] ) ) {
+                $email_addresses = $action_settings[ $setting ];
+            } else {
+                $email_addresses = explode( ',', $action_settings[ $setting ] );
+            }
+
+            // Loop over our email addresses.
+            foreach( $email_addresses as $email ) {
+                // Removes empty values that are passed in.
+                if( empty( $email ) ) continue;
```
kjohnson: This check does not account for spaces in the comma-separated list.
The value needs to be trimmed for whitespace before checked, otherwise empty merge tags result in a broken list.

For example:
`{wp:admin_email}, {field:email}` -> `kyle@ninjaforms.com, `

Should be:
`{wp:admin_email}, {field:email}` -> `kyle@ninjaforms.com`

ericwindham: @kbjohnson90 We pushed a change that should fix that issue.

kjohnson:
Closes #2974.

PR #3104, opened 2017-09-26
DO NOT RELEASE BEFORE ME (php 5.2)

Much2tall:
Broke up multiple function calls into separate lines for php 5.2 compatability.

@wpninjas/developers


klhall1987:
Tested and working properly. Merging into develop.

Issue #3015, opened 2017-08-30
Currency symbols in collect payment fixed amount is breaking payments


klhall1987:
If a user puts a currency symbol in their collect payment action's fixed amount, then it's causing the payment to fail. We probably need to poke-yoke this so user's can't enter a currency symbol here.

Issue #2886, opened 2017-07-19
Multi-Select Calculations are incorrect in front-end js


Much2tall:
Steps to reproduce:
- Create a new Form
- Add a Multi-Select Field
- Put that Field in a Calculation
- Add an HTML Field and put the Calculation in it

Expected behavior:
Selecting multiple items in the Multi-select should add them together. (1+2=3)

Current behavior:
Selecting multiple items concatenates them. (1+2=12)

PR #3106, opened 2017-09-27
Internationalization Update for Number Formatting and Currency

### LINKED ISSUE #2950: Locale Settings
- [ ] Store locale settings as Form data
-  [ ] Check backwards compatibility
- [ ] Extract locale mask to Helper class
- [ ] Create locale mask for js
-  [ ] builder
-  [ ] front-end
- [ ] Convert all stored number data in js to US locale
- [ ] Convert all saved data in subs to US locale
- [ ] Cross-check compatibility with Payment Gateway add-ons

Much2tall:
Fixes #2950 

- Addition of several per-Form settings to better manage international number formatting and currency symbols.
- Update of various controllers, both front-end and builder specific to better handle international calculations.
- Update to various processing methods to better handle international calculations.

@wpninjas/developers


Much2tall:
- [ ] In continued testing, Total Field was found to be non-functioning.

Much2tall:
Fixes https://github.com/wpninjas/ninja-forms/issues/2886

Much2tall:
Fixes #3015

klhall1987:
Right now all product fields append (decimal separator)00 at the to the end of each value that is input in the product field. Is this supposed to be the default behavior?

klhall1987:
When the number format has spaces in it ie(1 000 000,00), when it is displayed either commas or decimals are used as separators. It seems to be which ever one was used last.

klhall1987:
Here is a screenshot(http://nimb.ws/AdXjGR)

klhall1987:
I'm getting an NaN error in the total field, I switch from a number format that uses spaces as a thousands separator such as(1 000 000.00) to the number format that uses a decimal separator of the opposite type so when I switched from the one above to this number format(1.000.000,00). I was able to reliably reproduce the bug.  

Screenshot from the front end(http://nimb.ws/xLhqF2). 

Replication Steps:
1. Add a product and total field
2. Edit the product field and set the value to 1000
3. Go to the advanced tab and change the number format to 1 000 000.00. 
4. Publish and preview the form. 
5. Go back to editing the form and switch the number format to(1.000.000,00). 
6. Publish and preview.

klhall1987:
I think the NaN bug exists when switching between any two number formats that don't use the same decimal place separator. I just switch from 1,000,000.00 to 1.000.000,00 and got an NaN on the front end.

klhall1987:
Some what related to the other bug, if you switch between the two number formats and go back to edit the product field you have to completely remove separators from the product value before you can input new data. 

Here is a screen cast to better explain what I'm talking about to give you more detailed replication steps:
http://nimb.ws/UMFsjP

PR #3114, opened 2017-10-02
Issue#2951 2

ericwindham:
Fixes #

- This completes the 'Submission Count' merge tag

- We removed the previous NinjaForm code to simply use a query to pull the submission count

@wpninjas/developers


includes/MergeTags/Form.php
```diff
@@ -16,7 +18,8 @@ public function __construct()
 
         $this->merge_tags = Ninja_Forms()->config( 'MergeTagsForm' );
 
-        add_action( 'ninja_forms_save_sub', array( $this, 'setSubSeq' ) );
```
Much2tall: The scope of this PR was to get a sub count.
Why was the reference to sub sequence removed?

includes/MergeTags/Form.php
```diff
@@ -16,7 +18,8 @@ public function __construct()
...
+        // Gets the form ID.
+        add_action( 'nf_get_form_id', array( $this, 'set_form_id' ), 99999, 1 );
```
Much2tall: Although it's understood that this action should fire with priority, setting 99999 may not be a precedent we want to set, as that encourages us to take this shortcut again at later times.

Unless we're absolutely certain that this is the THE most important thing to fire, I would recommend cutting this back to 15 or so.

includes/MergeTags/Form.php
```diff
@@ -36,4 +39,34 @@ public function setSubSeq( $sub_id )
...
+     * @since 3.2.2
+     *
+     * @return array|int Count of the form submissions.
+     */
+    public function get_sub_count()
+    {
+		global $wpdb;
+
+		// Query the database for the total amount of submissions for a form.
+    	$query = "SELECT count(wpp.id) AS sub_count from wp_posts wpp JOIN wp_postmeta wpm ON wpp.id = wpm.post_id
```
Much2tall: Improper query escaping:
 - All tables should be prefixed (i.e. $wpdb->prefix . 'posts') This allows for databases with custom table prefixes. Otherwise, this will crash any site with non-standard prefixing in $wpdb.
 - All tables in query should be surrounded with \` characters, to avoid conflicts with older SQL versions. (i.e. \`table_name\`. This doesn't apply once you've converted the table to a named entity via AS, but any original table references need to be enclosed.
 - Code standards for SQL are that keywords are all caps. Query should read as follows:
SELECT COUNT(wpp.id) AS sub_count FROM \`wp_posts\` AS wpp JOIN \`wp_postmeta\` AS wpm ON wpp.id = wpm.post_id WHERE wpm.meta_key = '_form_id' AND wpm.meta_value = %s

includes/MergeTags/Form.php
ericwindham: It shouldn't have been. Accident.

includes/MergeTags/Form.php
ericwindham: Done

includes/MergeTags/Form.php
ericwindham: Done

ericwindham:
All requested changes added.

includes/MergeTags/Form.php
```diff
@@ -36,4 +41,36 @@ public function setSubSeq( $sub_id )
...
+     * @since 3.2.2
+     *
+     * @return array|int Count of the form submissions.
+     */
+    public function get_sub_count()
+    {
+		global $wpdb;
+
+		// Query the database for the total amount of submissions for a form.
+    	$query = "SELECT COUNT(wpp.id) AS sub_count FROM `" .
```
Much2tall: Should probably use SELECT DISTINCT( COUNT(wpp.id) ) to ensure that we don't get duplicates.

ericwindham:
Tested and works.

Issue #2942, opened 2017-08-01
The Word "of" in the Word/Character Count needs to be translatable.


jmcelhaney:
The Word "of" needs to be translatable.
https://github.com/wpninjas/ninja-forms/blob/c2cf72107a02764cac0f67367307611de72a7361/includes/Templates/field-input-limit.html

PR #3120, opened 2017-10-03
Resolve add new template syntax error

Much2tall:
- Moved trim of imports to inside of not is array check to ensure that we do not try to trim an array, breaking functionality.
- Added translatable text for the (of) in the input limit text.

@wpninjas/developers


Much2tall:
Fixes https://github.com/wpninjas/ninja-forms/issues/2942

PR #3121, opened 2017-10-04
Reset Calcs Collection on Reset of Field Collection

Much2tall:
- Added radio listener for reset of fieldCollection into calculations to prevent breakdown of merge tags when save progress is active.
 - Added string cast into decimal trim function to prevent uncaught errors of trim function call on a non-string value.

Fixes https://github.com/wpninjas/ninja-forms-save-progress/issues/68

@wpninjas/developers


klhall1987:
Everything works properly, we need to fix merge tags and then merge.

klhall1987:
Tested and working properly.

PR #3125, opened 2017-10-04
Removed hidden field type from the fields_table merge tag.

Much2tall:
@wpninjas/developers


klhall1987:
This looks fine and is likely good to merge. I need to talk with @wpnzach about this to make sure that we want to exclude hidden fields from this merge tag.

Much2tall:
@klhall1987 As a note, this only removes them from the {fields_table} merge tag. {all_fields_table} should still display them.

klhall1987:
@Much2tall that should be fine. I'm going to go a head and merge this then.

PR #3132, opened 2017-10-08
Remove trashed submissions from unique field check.

### LINKED ISSUE #3062: Bug with unique field in Form Restriction and deleted submissions
I already commented on a previous issue (#2911) but that issue is already closed so I may have to create a new one.

Basically, the code to check the uniqueness of the field (from MR #2941) does not take into account the `status` of the post (should be `publish`). If a submission has been deleted in the backend, it shouldn't be included in the query result.

In my opinion, a simple solution would be to update the SQL query to also check the `post_status` of the post in the table `$wpdb->prefix . 'posts'` (value of the field should be `publish`) using a `Join`.

## To reproduce the bug
1. Create a form and add a restriction on a field (eg: email field).
2. Submit the form from the frontend
3. In the backend, delete this submission
4. Submit the form again from the frontend, using the same email address as the one previously used.

## Expected result
The data should be submitted

## Versions
WordPress 4.8 with NinjaForms 3.1.9

Much2tall:
Fixes #3062 

 - Restricted unique field check to only published posts, to avoid trashed posts from blocking submissions.

@wpninjas/developers

### DIFF
--- a/includes/AJAX/Controllers/Submission.php
+++ b/includes/AJAX/Controllers/Submission.php
@@ -205,7 +205,7 @@ class NF_AJAX_Controllers_Submission extends NF_Abstracts_Controller
              */
             
             global $wpdb;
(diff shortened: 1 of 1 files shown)


klhall1987:
Tested and working properly.

PR #3137, opened 2017-10-10
Sorts options in ListCountry fields by name #3117

### LINKED ISSUE #3117: Country list - default sort
When adding a **country** field, the sort seems to be based on the value. 
In Dutch, "Belarus" is "Wit-Rusland"; yet it suddenly appears between countries with B.

isaactorresmichel:
Fixes #3117

- When adding a country field, the sort seems to be based on the value. This PR fixes that sorting ListCountry fields by country name.

@wpninjas/developers

### DIFF
--- a/includes/Fields/ListCountry.php
+++ b/includes/Fields/ListCountry.php
@@ -60,7 +60,11 @@ class NF_Fields_ListCountry extends NF_Abstracts_List
             $options[ $key ][ 'selected' ] = 1;
         }
 
-        return $options;
+	    uasort( $options, function ( $a, $b ) {
+		    return strcasecmp( $a['label'], $b['label'] );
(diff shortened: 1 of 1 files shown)


klhall1987:
@isaactorresmichel This PR is causing my site to Crash in PHP 5.2. Here is the error that is being thrown. 

Parse error: syntax error, unexpected T_FUNCTION in /app/public/wp-content/plugins/ninja-forms/includes/Fields/ListCountry.php on line 63

isaactorresmichel:
A fix was added to support legacy PHP env

@wpninjas/developers

jbostoen:
Thanks for the quick fix!
Will definitely look into supporting Ninja Forms! :)

PR #3139, opened 2017-10-11
Calcs Merge Tag not displayed after submission

Much2tall:
 - Modified set check to define default decimal places if the option is left empty. The original check was not properly catching blank values.

@wpninjas/developers


klhall1987:
Tested and working properly.

PR #3155, opened 2017-10-18
Add ability to display the Form ID and Form Title on the Submissions …

isleshocky77:
Add the ability to display the _Form ID_ and the _Form Title_ in the Submissions table.

*Note* I currently did not add this data to _Export_ and _Download All_ as the column filters does not currently change the output of this data. Adding the new data may cause issues with people who have templates set up for consuming this data.  This would need a larger conversation.

@wpninjas/developers

### DIFF
--- a/includes/Admin/CPT/Submission.php
+++ b/includes/Admin/CPT/Submission.php
@@ -163,6 +163,10 @@ class NF_Admin_CPT_Submission
 
         $columns['sub_date'] = __( 'Date', 'ninja-forms' );
 
+        $columns['form_id'] = __( 'Form ID', 'ninja-forms' );
+
+        $columns['form_title'] = __( 'Form Title', 'ninja-forms' );
+
         return $columns;
     }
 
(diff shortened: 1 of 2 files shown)


klhall1987:
@isleshocky77, 

Would you mind to clarify what issue this PR is trying to solve? Currently users have to select the form they would like to see the submissions of and if the users edits a form they will see the form name there as well(http://nimb.ws/BcVIxP).

isleshocky77:
It just adds an option for one of the columns shown to be the Form ID or Form Title. It helps if you're copy/pasting from the data on the screen. It would be more helpful to be in the export and download file; but I'm worried about effecting people's process if a new field comes out of no where.

Much2tall:
Closing PR without merging due to the following:

 - Submissions table to be re-addressed by the development team at a later date, potentially invalidating changes.
 - PR supports an edge use case, and could potentially create more problems than it resolves.

PR #3156, opened 2017-10-19
Replaced 'display:none' with 'type=hidden' in Hidden Field Type template

### LINKED ISSUE #3085: Hidden Field Template doesn't use type hidden
The field template for the Hidden Field Type does not use type `hidden`, but rather hard codes the inline style `display:none`. This can cause conflicts with themes/plugins/custom css in which the hidden fields are shown.

mscalessio:
Fixes #3085 

- Replaced hard coded inline style`style="display: none;"` with `type="hidden"` for Hidden Field Type

@wpninjas/developers


klhall1987:
Tested and appears to be working. Merging into the develop branch. @mscalessio Thanks for the PR.

PR #3171, opened 2017-10-27
Correct Number Field Step Errors

### LINKED ISSUE #3170: Number field incrementation error
Reproduction:

1. Add a Number field to a form, under Number Options setting, set Step to .01
2. In form preview (or with appended form) enter a value between 1047.57 - 1047.62. Error message Please increment by 0.01 appears

It looks like this 5 step range is the only numbers that throw this error, though there may be more. 1047.56 and below, or 1047.63 and above do not error.

Much2tall:
Fixes #3170 

- Modified type conversion to better handle rounding of numbers for modulus operations.

@wpninjas/developers


klhall1987:
Merging into develop.

PR #3191, opened 2017-11-07
Added an escape to remove place holders from the where clause.

### LINKED ISSUE #3182: Search in Submissions is not working in WP 4.8.3
The Search in Submissions table is not working since updating to WordPress 4.8.3

https://github.com/wpninjas/ninja-forms/blob/a8659baf2eff79852ad9b9d4b23be382748e129c/includes/Admin/Menus/Submissions.php#L208

klhall1987:
Fixes #3182 

- Added an escape to remove place holders(%) from the "where" clause. I wanted another set of eyes on this to see if this fix makes sense. I'm not sure if this fix could break anything else or not. We will also need to fix this in the 2.9.x codebase if approved. 

@wpninjas/developers

### DIFF
--- a/deprecated/classes/subs-cpt.php
+++ b/deprecated/classes/subs-cpt.php
@@ -598,13 +598,16 @@ class NF_Subs_CPT {
 
 			if (!empty($query)) {
 				// add to where clause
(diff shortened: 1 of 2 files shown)


ericwindham:
I've updated this to work for 2.9, also. Need KR or Kenny to confirm.

PR #3222, opened 2017-11-29
Remove null data from being sent everytime the db is updated

### LINKED ISSUE #3209: Do not send null data on form publish
Do not send data when the form is publish.

klhall1987:
Fixes #3209 

- Removes any null data from being sent to the database any time a change is made on the form. 

@wpninjas/developers


includes/Admin/Menus/Forms.php
```diff
@@ -360,6 +359,17 @@ private function _localize_form_data( $form_id )
         <?php
     }
 
+    private function null_data_check( $settings )
+    {
+        foreach ($settings as $key => $setting) {
+            if ( '' === $setting || null === $setting ) {
```
Much2tall: We only need a null compare here, as an empty string can be a valid input.

Issue #3151, opened 2017-10-17
Form import/export not working


TonyC5:
- **What actions are you taking?**
  - Exported form on one website (Website A) and attempted to import the form on another website (Website B).  Both websites running WordPress 4.8.2, Ninja Forms 3.2.2, PHP 5.6.31.
- **What are you expecting to occur?**
  - Expected duplicate form to be available after importing on Website B.
- **What is occurring instead?**
  - Fields of Form on Website B are not available in form editor, form shortcode does not work.  Actions were imported correctly (and the Save Form action was added as described in an earlier reported issue).
The imported form is unusable and uneditable.

Unfortunately, this forum does not allow me to attach the export file, since it is an unsupported type.  If renaming the file to a supported type works, or there is some other way you want me to send the exported form, let me know.

Much2tall:
UTF8 encoding issue.
Needs some serious dev discussion.

TonyC5:
@Much2tall - I identified another import/export issue and am including it here for completeness.  Not sure if this should be part of your import/export refactoring.

I imported two forms from my development webserver to my production webserver (both servers running WP 4.8.2, Ninja Forms 3.2.1, PHP 5.6). When I inspected the settings of the imported forms (on the production webserver), each form had a new ‘Save Form’ Action that I did not create and that is not present on my development webserver.

My forms each have a single action (Success Message). I’m implementing my form’s main action in the ninja_forms_run_action_settings Action Filter which is why my form has only the Success Message Action.

PR #3227, opened 2017-12-05
Issue#3151

ericwindham:
Fixes Issue #3151

This allows users to turn on/off UTF-8 encoding when importing/exporting forms. Some users were seeing blank forms when importing with the default utf-8 decoding.

@wpninjas/developers


klhall1987:
Tested and working. Merging into develop.

PR #3233, opened 2017-12-07
Remove JS error collecting. 

Much2tall:
- Commented out js error check to prevent repeated AJAX calls from bogging down server resources.

@wpninjas/developers


ericwindham:
Looks Good to me

PR #3235, opened 2017-12-07
Hide Credit Card Fields

Much2tall:
- Removed credit card field from field listings and added an array of fields to be excluded from the field search.
- Updated zIndex of jBox to make it properly show in the builder when a notice is registered.

@wpninjas/developers


klhall1987:
We need to add a bit of padding to the right side of the screen. Currently on a smaller screen size the box covers the Plus sign to add new fields and it's also half way off the page. 
http://nimb.ws/OkhjPs

Much2tall:
Made changes to this PR to satisfy review of the PR that relies on it.

PR #3257, opened 2017-12-18
Fix submission sequence on duplicate forms

### LINKED ISSUE #3136: Submission sequence number not incrementing on duplicated form
Few days ago I realized, that each submission in my forms, that were **duplicated** from a template form, has **sequence number 1** - it **does not increment**.

When I quickly looked at source codes, it looks like it is caused by the changes in method **duplicate()** from class **NF_Database_Models_Form**. It is also connected with how **get_next_sub_seq()** method in the same class works.

Much2tall:
Fixes #3136 

-Modified duplicaitonquery to not create a _seq_num record, rather than setting it to 0 by default.

@wpninjas/developers 

**Note:** This prevents the issue from occurring on new duplicates. Forms already experiencing this issue will not be fixed by this change.


klhall1987:
Tested and working properly. Merging into develop.

jbostoen:
Is there any update on this please?

RobbyButler:
The opening comment above said:

Note: This prevents the issue from occurring on new duplicates. Forms already experiencing this issue will not be fixed by this change.

However I just duplicated a form that had this problem and found I have the same problem with the new form. Apparently the only way to fix the old form is to rebuild it from scratch by hand. Correct?

If "Duplicate Form" were to fix the problem, it would then be simple to correct all my forms that have the problem by simply duplicating them and using the duplicate.

Thanks!

jbostoen:
Good question, or are there manual instructions to easily fix forms which are impacted?  
A query we can run or something?

PR #3260, opened 2017-12-20
Tide PHP Compatibility Updates

Much2tall:
- Resolved several instances of non-compatibility with various PHP versions, as reported by WPTide.

@wpninjas/developers


Much2tall:
One additional error not addressed in deprecated/includes/admin/ajax.php on line 541.

> Function each() is deprecated since PHP 7.2; Use a foreach loop instead.

klhall1987:
Tested and working properly.

PR #3262, opened 2017-12-20
Add checkbox value settings. 

### LINKED ISSUE #3076: Checkbox field ignoring custom values set in the field.
Checkbox field ignoring custom values set in the field. 

To replicate:
Create a form with a single checkbox
Go under the Advanced settings of the field and set values
Submit the form then check the submissions and the custom value that is set is not recorded

klhall1987:
Fixes #3076

- Commented all of the checkbox field file. 
- Added a copy of the field model to be passed in to the 'ninja_forms_subs_export_field_value_' filter. 
- Created settings for checked and unchecked values for the checkbox field. 

@wpninjas/developers

### DIFF
--- a/includes/Config/FieldSettings.php
+++ b/includes/Config/FieldSettings.php
@@ -144,6 +144,33 @@ return apply_filters( 'ninja_forms_field_settings', array(
     ),
 
     /*
+    * Checkbox Values
+    */
+    'checkbox_values' => array(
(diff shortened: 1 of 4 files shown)


klhall1987:
Added requested fixes, I will make issues for all the other add-ons to update their code once this has been  merged into develop.

PR #3264, opened 2017-12-22
fixes issue#3115 fields with currency masks not saving

ericwindham:
Fixes issue#3115

Saves text fields with 'currency' mask to the database correctly

@wpninjas/developers


klhall1987:
Working properly, merging into develop.

PR #3265, opened 2017-12-26
Remove broken handling of invalid date filters

shadyvb:
Invalid date filters shouldn't be handled by code, and the comparison between date strings doesn't work, as they're not integers.

Remove the extra handling of the next day, WP_Date_Query instead has an `inclusive` param that achieves the required behavior.

### DIFF
--- a/includes/Admin/Menus/Submissions.php
+++ b/includes/Admin/Menus/Submissions.php
@@ -426,7 +426,7 @@ final class NF_Admin_Menus_Submissions extends NF_Abstracts_Submenu
 
         $begin_date = $_GET[ 'begin_date' ];
         $end_date = $_GET[ 'end_date' ];
-        
+
         // Include submissions on the end_date.
         $end_date = date( 'm/d/Y', strtotime( '+1 day', strtotime( $end_date ) ) );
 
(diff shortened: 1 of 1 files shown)


includes/Admin/Menus/Submissions.php
```diff
@@ -427,20 +427,12 @@ private function table_filter_by_date( $vars )
         $begin_date = $_GET[ 'begin_date' ];
         $end_date = $_GET[ 'end_date' ];
 
-        // Include submissions on the end_date.
-        $end_date = date( 'm/d/Y', strtotime( '+1 day', strtotime( $end_date ) ) );
```
Much2tall: This line still needs to be included.

includes/Admin/Menus/Submissions.php
```diff
@@ -427,20 +427,12 @@ private function table_filter_by_date( $vars )
...
-            $end_date = $temp_date;
-        }
-
         if ( ! isset ( $vars['date_query'] ) ) {
 
             $vars['date_query'] = array(
                 'after' => $begin_date,
-                'before' => $end_date
+                'before' => $end_date,
+                'inclusive' => true,
```
Much2tall: The changes here don't appear to affect the filter.
'inclusive' is failing to do the same thing our original line 431 did.
Revert params to original settings.

shadyvb: That's true, reverted 👍

PR #3276, opened 2018-01-03
Feature/all filled fields mergetag

ericwindham:
This adds the 'all_filled_fields_table' merge tag

It is similar to the 'fields_table' merge tag as it ignores empty fields(not filled in by the user). But, it does include html and hidden fields that the 'fields_table' does not.

@wpninjas/developers


ericwindham:
After discussions with Zach, we are not continuing with this Merge Tag since the {fields_table} accomplishes the same goal.

Eric Windham

PR #3302, opened 2018-01-19
Decimal setting not honoring 0

### LINKED ISSUE #3224: Calculation decimal setting not respected in success message
To Reproduce:

1. Create calculation, set decimal setting to 0
2. Include calculation in success message action
3. Submit form, view calculation. Value will display as xx.xx instead of just xx

Much2tall:
Fixes #3224 

- Removed unnecessary check that was causing a 0 value in the decimal place setting to resolve as the default value of 2 instead.

@wpninjas/developers


klhall1987:
Tested and working merging into develop

PR #3303, opened 2018-01-20
Check if global post is set before using it

ywatt:
With some themes editors it can append that there is no post and the current code fires a few php warning like:
PHP Notice:  Undefined index: post in /home1/public_html/test_site/wp-content/plugins/ninja-forms/includes/Admin/Metaboxes/AppendAForm.php on line 18

Fixes #

- The proposed change is a defense against a non set post.
- Test first if it is set, otherwise set to NULL.

@wpninjas/developers


Much2tall:
@ywatt

ywatt:
@Much2tall thanks for the feedback. Regarding the ternary I checked first the coding rules and it mentions that it is ok as long as it is not negative condition.
I must admit that I actually prefer if/then/else as well :)
Regards

Issue #3334, opened 2018-02-02
Merge tags causes too many post meta queries


paulgibbs:
In the merge tags code in `NF_MergeTags_Deprecated` and `NF_MergeTags_WP`, which lets a user insert a post meta value into the post-submission screen (or post type), it fetches all post meta via an uncached SQL query. This is hooked to `the_post`.

The problem with this is encountered when a single post has lots and lots of post meta, or a single screen renders lots and lots of posts. I have a client site I am working on that meets both criteria (unfortunately), and this small bit of code is adding 0.4 seconds to page generation time for every single front-end screen.

PR #3335, opened 2018-02-02
Fetch post meta for merge tags on demand, rather than upfront.

paulgibbs:
Decreases memory use on posts with lots of post meta, and problems caused by some page builders triggering the `the_post` action multiple times to render one screen.

For #3334

- Fetch post meta on demand, rather than all up-front, for performance improvements (speed and memory).
- Remove an uncached SQL query.

@wpninjas/developers

### DIFF
--- a/includes/MergeTags/WP.php
+++ b/includes/MergeTags/WP.php
@@ -31,7 +31,7 @@ final class NF_MergeTags_WP extends NF_Abstracts_MergeTags
 
         /**
          * Replace Custom Post Meta
-         * {post_meta:_meta_title} --> meta key is 'foo'
+         * {post_meta:foo} --> meta key is 'foo'
          */
         preg_match_all( "/{post_meta:(.*?)}/", $subject, $post_meta_matches );
         if( ! empty( $post_meta_matches[0] ) ) {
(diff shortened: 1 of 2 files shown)


includes/MergeTags/WP.php
```diff
@@ -62,9 +40,11 @@ public function replace( $subject )
              * $matches[1][$i]  captured meta key   foo
              */
             foreach( $post_meta_matches[0] as $i => $search ) {
-                $meta_key = $post_meta_matches[1][$i];
-                if ( isset( $this->post_meta[ $meta_key ] ) ) {
-                    $subject = str_replace( $search, $this->post_meta[$meta_key], $subject );
+                $meta_key   = $post_meta_matches[1][$i];
+                $meta_value = get_post_meta( $this->post_id(), $meta_key, true  );
+
+                if ( $meta_value ) {
```
Much2tall: It's possible for the meta value to be a valid 0.
This check will cause any such values to be excluded.
Please update to check against an empty string instead.

Aside from that, very nice work!
It's hard to believe how much inefficient code we had running to do just this.

Much2tall:
@paulgibbs 
Please see review for requested changes.

PR #3359, opened 2018-02-14
fixes IE error of forms not displaying

### LINKED ISSUE #3358: Form not displaying in IE 11 in Ninja Forms 3.2.12
- **What actions are you taking?**
  - What had happened was... 
The form loads in all other browsers, but I visited a Ninja Form in Internet Explorer 11 on Ninja Forms 3.2.13 and the form did not load
- **What are you expecting to occur?**
  - What I was expecting to happen was...
The "Fields marked with an * are required" text appeared, but the form did not load below it
- **What is occurring instead?**
  - What *actually* happened was...
There is a console error related to autocomplete in the console.  It says "Object doesn't support property or method "includes".
https://goo.gl/rz9AEz

ericwindham:
Fixes #3358 

removes the '.includes' function on the array on line 11 and replaces it with a call to '.indexOf'.

'.includes' is not supported in IE

@wpninjas/developers


klhall1987:
Tested and working properly.

Issue #3000, opened 2017-08-21
Star Rating Default Value


kjohnson:
The Star Rating field's default value is currently used to determine how many stars to display.

This default value is submitted as the value when no other value is selected, but doesn't show as the value.

**Suggestion**: Update the `fieldStarRating.js` controller to reset the value after initial render.

---
Reference: https://wpninjas.kustomerapp.com/app/customers/599975e70ace03001103ef5f/event/5999941d3de45900111a37d5

kjohnson:
Related to #2815 as the number of stars is the default value.

PR #3370, opened 2018-02-16
Fixes Star Rating issues - Issue#2815

### LINKED ISSUE #2815: The Star Rating field is not being caught by the required field validation
The Star Rating field is not being caught by the required field validation in Ninja Forms 3.1.4

ericwindham:
Fixes #2815 and #3000 

Create a field with a star rating. It now has a default value, that when set, should check the appropriate number of starts. And when left blank, and the field is set to required, should trigger a required error on submit if the field is left blank.


klhall1987:
Tested, working, and merging develop

PR #3399, opened 2018-02-27
Ninja Forms Services

kjohnson:
- Adds a new "Services" section to the Dashboard.
- Adds a custom email header as a flag for transactional email.

@wpninjas/developers


klhall1987:
Hey @kbjohnson90, 

I did a few test emails this morning. A few things I've noticed. 

- I think Submissions to take longer on when sending things to feed blitz. I don't know if this covered in the scope of this pass or not. 
- HTML doesn't seem to be coming through in the email. I added an H1 tag in the body of the email and it didn't seem to come through. So that might be something you want to take a look at.

kjohnson:
Fixed the issue with HTML being escaped.

klhall1987:
@kjohnson I got a Fatal error on load after I switched to this branch on 5.2. We may need to only load the services tab, if they are above what ever the required PHP version for the Mailer add-on is. 


( ! ) Fatal error: Can't use function return value in write context in /app/public/wp-content/plugins/ninja-forms-recurly/includes/Actions/Subscribe.php on line 84
--

ericwindham:
Also, just a heads up. Some of the files aren't commented to the style guide standards.

ericwindham:
Getting a 403 error on my.ninjaforms.com when trying to authorize. 

{"code":"rest_cookie_invalid_nonce","message":"Cookie nonce is invalid","data":{"status":403}}

ericwindham:
As for the style guide. One of the main things is docblock style comments should precede any function. Most of the code is readable and understandable, but we just need to be super cognizant of the fact that others will more than likely be supporting the plug-in. So we don't need to be redundant in commenting, but right now(especially new functionality) can't really be over-commented. The function docblocks will go a long way.

PR #3411, opened 2018-03-01
deactivate and delete large-cache-fix plugin, implement it into core

ericwindham:
- This implements @kjohnson's large-cache-fix into core
https://github.com/kjohnson/nf-large-cache-fix

- So we have to deactivate it and delete it if it's installed
- Change the CHUNK_SIZE constant to a smaller number to test

@wpninjas/developers


ericwindham:
handling this a different way

PR #3413, opened 2018-03-01
Akismet Anti-Spam integration

### LINKED ISSUE #291: Integrate with Akismet for anti-spam
It would be nice to integrate with Akismet out of the box as another anti-spam measure.

polevaultweb:
Fixes #291

- Adds new action where you can select fields to be used as author, email, url and message to be checked against the Akismet service for spam.

@wpninjas/developers


klhall1987:
@polevaultweb, I'm not seeing the Akismet action. I've downloaded jetpack and akismet on my local, but I'm still not seeing the action in the action drawer. Any idea as to what is going on?

klhall1987:
Akismet doesn't validate if it is not connected to the service, which is fine, but we probably need to throw an error and let the user know they need to connect to Akismet to use the action.

polevaultweb:
@klhall1987 fixed up

klhall1987:
Seems to be working properly merging into develop.

PR #3418, opened 2018-03-05
deactivate and delete large cache fix

ericwindham:
This deactivates Kyle's large cache fix and deletes the files. We are going to implement something similar

https://github.com/kjohnson/nf-large-cache-fix

@wpninjas/developers


klhall1987:
Tested and working properly.

PR #3426, opened 2018-03-08
Handle chunked cache

ericwindham:
This adds cache chunking to the front-end builder and the backend. The default chunk size is set to 1000, so you may need to reduce that value(in both js and php backend).

@wpninjas/developers


klhall1987:
Tested as best as I can. Merging into develop.

PR #3445, opened 2018-03-22
Auto complete ready

ericwindham:
This adds autocomplete capabilities. Any field that uses a textbox on the front-end will now be able to 'disable autocomplete' and have a 'custom name attribute' in the advanced settings with a default value for the it's corresponding field type

@wpninjas/developers


klhall1987:
Tested and merging into develop.

PR #3447, opened 2018-03-22
Undo Telemetry 

Much2tall:
- Added controller for dispatcher end points. Added undo as the first method of this class.
- Added new controller to core file calls.
- Created function to be called when undo is used, creating a 1 week cookie to prevent further sends and sending a call to the undo end point.

@wpninjas/developers


klhall1987:
Tested and seems to be working. Please fix the merge conflict then you're free to merge the PR.

PR #3454, opened 2018-03-26
Added sanitization for list values upon update of the value.

### LINKED ISSUE #3014: Sanitize list values of special characters
Special characters in list values, such as parenthesis, break both conditions and calculations.  We need to sanitize the inputs or make them more robust to prevent issues in these areas.

Much2tall:
Closes #3014

@wpninjas/developers


klhall1987:
Looks to be working properly.

PR #3455, opened 2018-03-27
Updated sanitization to only fire on Field objects.

Much2tall:
@wpninjas/developers


klhall1987:
Tested and working properly.

PR #3465, opened 2018-04-02
Modified DOM constructor to used proper selection attribute.

Much2tall:
@wpninjas/developers


klhall1987:
Tested and seems to be working properly.

PR #3468, opened 2018-04-04
create unique nonce values for each form render

ericwindham:
This creates a unique nonce so that we can avoid users getting nonce values that could possibly expire before they submit their form.

@wpninjas/developers


assets/js/front-end/controllers/submit.js
```diff
@@ -32,7 +32,7 @@ define([], function() {
 		 * @return void
 		 */
 		submit: function( formModel ) {
-
+			console.log(nfFrontEnd);
```
Much2tall: Should probably remove this.

includes/AJAX/Controllers/Form.php
```diff
@@ -138,4 +139,22 @@ public function delete()
 
         $this->_respond();
     }
+
+	/**
+	 * Let's generate a unique nonce for each form render so that we don't get
+	 * caught with an expiring nonce accidentally and fail to allow a submission
```
Much2tall: `* @since 3.2`

includes/Display/Render.php
```diff
@@ -575,7 +575,7 @@ public static function enqueue_scripts( $form_id, $is_preview = false )
         wp_localize_script( 'nf-front-end', 'nfi18n', Ninja_Forms::config( 'i18nFrontEnd' ) );
 
         $data = apply_filters( 'ninja_forms_render_localize_script_data', array(
-            'ajaxNonce' => wp_create_nonce( 'ninja_forms_display_nonce' ),
+//            'ajaxNonce' => wp_create_nonce( 'ninja_forms_display_nonce' ),
```
Much2tall: Just remove this instead of commenting it.

assets/js/front-end/main.js
```diff
@@ -126,6 +128,42 @@ jQuery( document ).ready( function( $ ) {
 				}
 			},
 
+			/**
+			 * This function retrieves a new, unique nonce so that we avoid
+			 * giving the user a nonce that could possibly expire before
+			 * they finish filling out the form.
```
Much2tall: `* @since 3.2`

PR #3469, opened 2018-04-04
adds an action to access and act on $data before responding

ericwindham:
A necessary change to make fixing the following Save Progress issue doable:

https://github.com/wpninjas/ninja-forms-save-progress/issues/96

This change can be merged independently of any Save Progress changes for that issue

@wpninjas/developers


klhall1987:
Tested and working properly.

PR #3494, opened 2018-04-17
Fixes #3490 (Ninjaforms is not-usable for RTL users)

### LINKED ISSUE #3490: Ninjaforms is not-usable for RTL users
- **What actions are you taking?**
1. The user had selected an RTL language (Hebrew)
2. The user creating/editing a form
3. The user opens the "fields drawer"

  - What had happened was... 
The user can't see the fields

- **What are you expecting to occur?**
"Fields drawer" shouldn't hide the fields area

Solution:
In rtl.css file, add:
```
.nf-drawer-opened .nf-main-test {
   float: left;
}
```

AlmogBaku:
Fixes drawer alignment for RTL users

Fixes #3490

- mark the overriding in the `admin/rtl.scss` as `!important` so they won't be ignored by inline css.
- recompilation of the scss files

@wpninjas/developers


AlmogBaku:
Hiding of the drawer is broken... working on a fix

PR #3509, opened 2018-04-25
Correct busted field keys at generation.

### LINKED ISSUE #2552: Foreign characters for Labels cause Field Keys to generate improperly
(Japanese characters are the known offender)

Steps to replicate:
- Create a new Form
- Add a new Field
- Change its Label to all Japanese characters
- Publish the Form
The added Field will be missing from the preview.
If you go back into the Builder and check the Field Key for that Field, it will be only numbers, with no text or underscore.

Much2tall:
Fixes #2552 

- Added check to force valid field keys when they are generated improperly, such as due to foreign character input.

@wpninjas/developers


AlmogBaku:
I can verify that the fix on develop work

PR #3540, opened 2018-05-09
Renamed modal-ad to modal-template

wpnzach:
Because "ad" was in div class name, it was being hidden by ad blockers.

Fixes #

-
-
-

@wpninjas/developers


ericwindham:
Looks good

PR #3542, opened 2018-05-10
More accessibility changes

ericwindham:
Create a form with all type of fields. Check for 'aria-labelledby' attributes on field elements and 'required' attributes on field elements of fields set to required.

@wpninjas/developers


klhall1987:
Tested and working properly.

PR #3559, opened 2018-05-16
Gdpr data request data removal

ericwindham:
You'll need the WordPress Beta Tester Plugin to test with the new Erase/Export user data functionality.

See the new templates, submit to a form using each template. Then, check the corresponding pages under 'Tools' in the WordPress admin panel.

Under the 'email' field in each table click on the action and make sure they work. Any submissions should be erased from the post and postmeta tables when using the erase data functionality, and you should see submission data on the HTML exported page when using the export data functionality.

@wpninjas/developers


klhall1987:
Working properly.

PR #3562, opened 2018-05-17
Gdpr optin

Much2tall:
See commit log for changes.

@wpninjas/developers


client/dashboard/views/dashboardView.js
```diff
@@ -93,6 +92,155 @@ define( [ 'views/sections/widgets.js', 'views/sections/apps.js', 'views/sections
...
+                var titleStyling = document.createElement( 'h2' );
+                titleStyling.style.fontSize = '130%';
+                titleStyling.innerHTML = 'Help make Ninja Forms better!';
+                titleStyling.style.textAlign = 'left';
+                title.appendChild( titleStyling );
+                // Define the modal content.
+                var content = document.createElement( 'div' );
+                content.style.padding = '5px 15px';
+                var p = document.createElement( 'p' );
+                p.innerHTML = 'We would like to collect data about how Ninja Forms is used so that we can improve the experience for everyone. This data will not include ANY submission data or personally identifiable information.';
```
kstover: If we hard code this text string into the JS, can it be translated?

kstover:
We should also investigate just having the spinner replace the text of the button, rather than making the buttons disappear and a spinner be revealed. As it is, it conveys that something is happening, but it feels a bit jarring.

klhall1987:
Seems to be working properly. Waiting on KR to make a few other changes before merging.

PR #3591, opened 2018-05-30
Anonymize data

ericwindham:
Lot of changes here

Need to test with submission data from multiple users. Fill out forms with email, name, and address fields while logged out. Put an actual user's email address in the email field while logged out to test everything.

@wpninjas/developers


includes/Admin/UserDataRequests.php
```diff
@@ -54,69 +81,66 @@ function plugin_user_data_exporter( $email_address, $page = 1 ) {
...
+
+				foreach ( $fields as $field_id => $field ) {
+					// we don't care about submit, hr, divider, html fields
+					if ( ! in_array( $field->get_setting( 'type' ),
+							$this->ignored_field_types ) ) {
+						// make sure there is a value
+						if ( isset( $sub_meta[ '_field_' . $field_id ] ) ) {
+
+							//listcheckbox fields may need to be unserialized
+							if( 'listcheckbox' == $field->get_setting( 'type' ) ) {
```
Much2tall: Should probably check multiselect here too.

includes/Admin/UserDataRequests.php
```diff
@@ -147,38 +172,252 @@ function plugin_user_data_eraser( $email_address, $page = 1 ) {
...
-			'items_retained' => $items_retained,
-			'messages'       => $messages,
-			'done'           => true,
-		);
+	/**
+	 * This will (redact) personal data and anonymize submissions
+	 *
+	 * @param $subs
+	 */
+	private function anonymize_submissions( $subs ) {
```
Much2tall: This files need comments from this point down.

